PR #12, opened 2023-04-29
Restructure content and rework sidebar

danisztls:
Fix #7 

### Reorganized content

Moved posts to a single directory removing sections there. Sections and pages have a one-to-many relation. Taxonomies and pages have a many-to-many relation. As we are using taxonomies there it doesn't make sense to have sections. But yeah, this ain't an rule.

e.g  "Verge suffers a 51% attack in April 2018" was under the "51%" section but had both "Protocol exploit" and "51%" categories.

### Sidebar

Improved the separation between the different kinds of content/links.

Screenshot from 2023-04-29 00-43-54

### Post meta

Added the taxonomy name before the terms list.

Screenshot from 2023-04-29 00-47-35

### Terms pages

Created a tag list layout for terms pages.

Screenshot from 2023-04-29 01-08-21

PR #16, opened 2023-05-13
Implement collapsibles on sidebar menu

danisztls:
Follow up #12

### DIFF
--- a/assets/_collapsible.scss
+++ b/assets/_collapsible.scss
@@ -1,0 +1,35 @@
+.collapsible {
+  // control containers
+  h2, h3, h4, h5, h6 {
+
+    display: grid;
+    grid: auto / auto max-content;
+    align-items: center;
+  }
+
+  .collapsible-control {
+    background: none;
+    border: none;
+    outline: none;
+    cursor: pointer;
+    font-size: 1rem;
+    font-weight: 300;
+
+    &:hover {
+      font-weight: 400;
+    }
+
+    &[aria-expanded="false"]:after {
+      // content: "";
+      content: "";
+    }
+
+    &[aria-expanded="true"]:after {
+      // content: "";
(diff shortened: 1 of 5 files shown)


marina-chibizova:
`Custodians` may extend dramatically in the future. let's make sure they are not cluttering the outline and the list only appears on the page opening by click. The same may be applied to `Categories`

danisztls:
@marina-chibizova What I have in mind is to have the level 4 sections (`Categories` & `Custodians`) collapsed by default but expanded if they are part of the current page path.

Like this:

Screenshot from 2023-05-15 07-41-25
Screenshot from 2023-05-15 07-41-38

marina-chibizova:
It's good, but we may have hundreds of custodians at some point. even as a collapsible, it will be too much for a menu. I would instead show this taxonomy on a page

danisztls:
Ok. Will write a rule to not list sections with more than 7 items.

marina-chibizova:
why 7? maybe just hide sections on this level of menu (I believe it's 4th) and only show higher level menu items? and list the details on pages. wouldn't it be easier and more straightforward logic?

danisztls:
> why 7?

No special reason. Can be 3 or any number.

> maybe just hide sections on this level of menu (I believe it's 4th) and only show higher level menu items? and list the details on pages. wouldn't it be easier and more straightforward logic?

Both ways are straightforward. I can comment the terms subsection or add a conditional statement. "Custodians" & "Categories" are on the same level. Wasn't an initial goal to have "Categories" expanded?

marina-chibizova:
I believe it's ok to not have `Categories` expanded, since we have them on the page + in the ToC on the right-hand side. No need to duplicate it in the main menu

marina-chibizova (approved):

PR #18, opened 2023-05-13
Fix comical font

### LINKED ISSUE #17: Fix font
after the last pr website is acting strange in mobile view and just on the web:

086b4cee-4acf-451b-a80f-d844c19f2cdc

Could you take a look please?

danisztls:
Fix #17

_Note: Comical as in "Comic Sans"_

## What happened

I implemented icons using webfonts. The icon fonts are fallback fonts. So when the main font doesn't contain the glyph the icon font is used. The problem is that the theme isn't loading the Roboto font as I expected and I didn't noticed that because I have those fonts installed locally.

## Fix

Moving `sans-serif` to be preferred earlier will make browsers use their system default sans-serif font instead of the comical "Font Awesome" font.

### DIFF
--- a/assets/_fonts.scss
+++ b/assets/_fonts.scss
@@ -40,5 +40,5 @@ $font-path: "/fonts";
 }
 
 body {
-  font-family: "Roboto", "Font Awesome 6 Sharp", "Font Awesome 6 Brands", sans-serif;
+  font-family: "Roboto", sans-serif, "Font Awesome 6 Sharp", "Font Awesome 6 Brands";
 }
--- a/assets/_fonts.scss
+++ b/assets/_fonts.scss
@@ -1,11 +1,40 @@
 $font-path: "/fonts";
 
 /*!
+ * Google Fonts
+ */
+
+@font-face {
+  font-family: 'Roboto';
+  font-style: normal;
+  font-weight: 400;
+  font-display: swap;
+  src: local(''),
+       url('#{$font-path}/roboto-v27-latin-regular.woff2') format('woff2'),
+}
+@font-face {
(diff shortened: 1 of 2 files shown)


marina-chibizova:
can you just go back to what it used to be before the pr? let's not over complicate the fonts

danisztls:
> can you just go back to what it used to be before the pr? let's not over complicate the fonts

It's using the same font as before. "Roboto" for text, "Roboto Mono" for code and "sans-serif" as fallback.

The icon font is needed so to use icons like the Github icon and the collapse/expand buttons of the menu sections.

Screenshot from 2023-05-14 09-06-49

assets/_fonts.scss
```diff
@@ -1,12 +1,41 @@
 $font-path: "/fonts";
 
+/*!
```
danisztls, line 3: @marina-chibizova While I'm adding a bunch of lines here they are actually carried over from the theme respective [file](https://github.com/alex-shpak/hugo-book/blob/d86d5e70c7c0d787675b13d9aee749c1a8b34776/assets/_fonts.scss).

marina-chibizova: makes sense. I would change the collapsible icon though, to just an arrow without the circle

danisztls: > to just an arrow without the circle

ok

danisztls:
@marina-chibizova Is there anything else?

marina-chibizova (approved):

marina-chibizova:
not really, let's see how it looks

marina-chibizova:
looks good, at least on my devices. thanks!

PR #23, opened 2023-05-22
Fix content path

danisztls:
There was a trailing space in the directory name.


danisztls:
Suggestion: remove the post date from the file names because Hugo don't use.

evgenydmitriev:
Why are you deleting `content/attacks/posts/2019-01-05-Ethereum-Classic.md`?

danisztls:
@evgenydmitriev I thought it was an update meant to replace the older post but I was wrong.

danisztls:
Rebased, restoring the deleted file.

PR #24, opened 2023-05-22
Integrate fact checking action

### LINKED ISSUE #14: Integrate action
[Here](https://github.com/1712n/challenge/issues/96#issuecomment-1540179265) is a solution that we need to integrate on this [website](https://dn.institute/). The solution is good, but I would appreciate your optimizations (e.g. he is using his personal token instead of a bot)

danisztls:
Fix #14


danisztls:
GitGuardian warning is a false positive

marina-chibizova:
@alxarno I will need your help with review

danisztls:
The script is failing because of:

`Rate limit reached for default-gpt-3.5-turbo in organization org-qifYVoOOQvrz24XJJ6GjJmpb on requests per min. Limit: 3 / min. Please try again in 20s. Contact us through our help center at help.openai.com if you continue to have issues. Please add a payment method to your account to increase your rate limit. Visit https://platform.openai.com/account/billing to add a payment method.`

marina-chibizova:
@danisztls do you need help with this?

marina-chibizova:
@danisztls done

danisztls:
@marina-chibizova This is the action output. Is this what is expected?

marina-chibizova (approved):

marina-chibizova:
yes, let's merge this one. you can also add some caching to not process each commit every time - something like this https://github.com/probot/metadata. but let's leave it for next pr

danisztls:
@marina-chibizova

> let's merge this one

A nitpick but I recommend squashing instead of merging, the result will be the same but the Git log will make more sense. 😄 

> you can also add some caching to not process each commit every time - something like this https://github.com/probot/metadata.

It doesn't process each commit every time. It process files modified by the PR when it's marked as ready for review.

I would cache the workflow metadata in the repo itself instead of in the PR as a comment like `probot/metadata` do. It might be useful to have a fact checking metadata file for each content file. Though caching might not be that useful as the model isn't so deterministic and it can extract/generate similar statements with different wordings in each run.

Currently there's kind of a bug where in the first run it extracts and verifies a few statements and in subsequent runs it verify a bit more. Last time it was exceptionally triggered twice due to the CODEOWNERS change and in the second run it verified an additional statement which was identified as false but yet that is covering ~half of the content only. What's is happening is that the model is timing out as it normally does. I should debug and fix this, probably by chunking the prompts in smaller bits so it doesn't time out.

PR #48, opened 2023-06-01
Update 2021-08-10-poly-network.md

evgenydmitriev:


evgenydmitriev:
@alxarno at the moment, the fact checking is running on the file being modified, instead of the additions.

PR #106, opened 2023-06-08
Rearranging the articles' text!

TheT1meWizard:
Hello! I'm sorta newb here and just wanted to apply for a vacancy at hh.ru. 

I've inspected some of your data and I felt that it's really hard to read due to weird articles' structure.
So, I've created a small [script](https://github.com/TheT1meWizard/your-wiki-fixer/blob/main/the-wiki-fixer.py) to rearrange articles' text in a more logical way: 
1. Summary
2. Timeline
3. Attackers
4. Losses
5. Security failure problems
6. Additional detail
I feel like this order gives a more readable flow of information and is more pleasant to the viewer. We get summary, followed by timeline of events, then we learn about attackers (and their methods). After that we see the damage done. Additional details are optional so they're in the very bottom.

Hope you have a good day!

Sincerely,
TheT1meWizard


evgenydmitriev:
Thank you for your submission, but closing, as this is not something that adds value to the current structure.

PR #142, opened 2023-06-27
Restructure  repo scripts

danisztls:
Overall goal is to facilitate tools improvement, maintenance and review.

- Implement a single package modular structure to facilitate code reuse and review.

- Get the PR diff using a different method which requires only the PR URL as input. This is simpler and allow contributors to run and test the script locally. GitHub environment is required only to comment on issue.

- Improve logging via decorators and GH logging groups.

- Misc. refactoring.

- Development mode produces poor results but is quicker. To use it include the optional parameter `--mode development`.


alxarno (approved):
Looks good

PR #117, opened 2023-06-18
Taxonomies: Improve pagination & layout

danisztls:
Fix #36

### Changes

- Order terms in the book section *(on the right side)* by their size *(n of pages)*
- Order terms in terms list alphabetically
- Increase pagination size to 50 in terms list
- Implement new layout where terms are grouped by their initial letter *(drop cap)*

PR #158, opened 2023-06-29
Fix taxonomies metadata

danisztls:
Follow up #117

Apparently there is no order of preference in config cascade so rules are applied randomly resulting in the taxonomy page sometimes rendering with the wrong template. But metadata in front matter has preference over cascade which solves the issue.


evgenydmitriev (approved):

evgenydmitriev:
@marina-chibizova, ready for the switch?

PR #167, opened 2023-07-04
Change headers

marina-chibizova:
making headers more generic and broad, to allow categorization based on entity name, type or type of attack. After approval I will fix headers in existing posts. @danisztls please check that it will not ruin current taxonomy

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,10 +3,16 @@
 Before committing to the wikis, please ensure your submission meets the following criteria:
 
 - File name - YYYY-MM-DD-entity-that-was-hacked.md
-- Required headers:
-  - Date: YYYY-MM-DD
-  - Categories: "Type of attack"
-  - Title: "Article Title"
+- Headers:
+
+| Field | Required | Description | Example |
+|--|--|--|--|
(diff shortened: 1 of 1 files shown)


JediFaust (approved):
I agree that making the headers more generic and broad is a great idea. This will indeed offer more flexibility for categorizing based on the entity name, type and type of attack.

JediFaust:
@danisztls Please, let me know when to start requesting headers to match the last changes

danisztls:
@marina-chibizova It looks good.

@JediFaust The new taxonomies will be in effect after #169 merge.

danisztls:
Not sure if this can be useful but Hugo has a template feature called [Archetypes](https://gohugo.io/content-management/archetypes/). I don't think content creators will be using running Hugo locally to benefit from `hugo new` command but they might still benefit from a file that they can copy and use as scaffold.

marina-chibizova:
> they might still benefit from a file that they can copy and use as scaffold.

sure, sounds great. it's something I would like to leverage, as people often struggle with the structure

PR #169, opened 2023-07-04
Refactor headers

marina-chibizova:
@danisztls I refactored the headers according to your last changes, but I'm not sure I got the logic right. Feel free to fix it and add missing parts so that the taxonomy works.

@JediFaust I used my best judgment to map types of attacks and entities. I would really appreciate your review and suggestions. (i.e. I'm not always sure about the right entity type - Blockchain or Coin. Everything DeFi I just mark 'DeFi' for simplicity and so on). What would be valuable and actionable taxonomy categories/tags for the community to use? Both universal and specific enough.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-## Wiki PR Approval Criteria
+## Attack Wiki PR Approval Criteria
 
 Before committing to the wikis, please ensure your submission meets the following criteria:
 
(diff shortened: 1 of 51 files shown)


danisztls (changes requested):
Some files still have the Custodians taxonomy.

- 2021-04-13-Africrypt.md
- 2021-12-04-Bitmart-Exchange.md
- 2023-03-13-Euler-Finance.md
- 2022-11-01-Deribit.md

marina-chibizova:
@danisztls thanks, fixed them. Do I understand this right, that content creators should use `attacks/posts/target-entities`, `attacks/posts/entity-types`, `attacks/posts/attack-types` as headers from now on? (before we implement templates for them to fill) In this case we need to announce it to reviewers and change documentation for content creators

danisztls:
> Do I understand this right, that content creators should use attacks/posts/target-entities, attacks/posts/entity-types, attacks/posts/attack-types as headers from now on? (before we implement templates for them to fill) In this case we need to announce it to reviewers and change documentation for content creators

@marina-chibizova Yes if we want to have taxonomies that are section-wide instead of site-wide. But `attacks/posts/attack-types`, `attacks/attack-types` and `attack-types` are all both valid ways of organizing content.

danisztls (approved):
Looks good

marina-chibizova:
> attacks/posts/attack-types, attacks/attack-types and attack-types are all both valid ways of organizing content.

then what was the reason to rename some headers in one of the previous PRs? let's decide on the one standard header format here so it won't confuse anyone

danisztls:
@marina-chibizova So the taxonomy is specific to the section `attacks/posts/` as it implied by the menu structure. As I understood from previous discussion we might have different sections with similar taxonomies at a later point. This avoids conflating `attacks/posts/entity-types` with `something-else/posts/entity-types`. Is this still pertinant?

marina-chibizova:
@danisztls I don't think we'll have the same taxonomy names, especially now when they are more specific to attacks. My concern is that not everything was shown on the website, like the file I mentioned - content/attacks/posts/2019-11-27-Upbit.md

danisztls:
@marina-chibizova 

### Why the taxonomy rename

> what was the reason to rename some headers in one of the previous PRs?

Expanding on this, my motivation was to fix an UX issue and this was the best workaround I found.

e.g.

In the page "Custodians" which is a term of the taxonomy "Categories" its parent sections are in bold in the left menu.

Screenshot from 2023-07-05 14-51-28

The URL path is used to infer and highlight the current page and its parent sections. But as the path of the taxonomy is its name it wasn't working properly because this term was located at `/categories/custodians` and not at `/attacks/posts/categories/custodians` and thus not int the expected path.

*Note that the right sidebar needs to be worked*.

### Pages not show

> My concern is that not everything was shown on the website, like the file I mentioned - content/attacks/posts/2019-11-27-Upbit.md

That file and a few others are using the old taxonomies `custodians` and `categories` instead of the new `attacks/posts/custodians` and `attacks/posts/categories`.  

You fixed that in this PR. But as that will happen again gonna create an automation to fix it automatically for future PR merges *(it's trivial)*.

marina-chibizova:
> But as that will happen again gonna create an automation to fix it automatically for future PR merges (it's trivial).

sure, that's gonna be the next step after we decide on the header format, merge this pr and announce the new rules

> But attacks/posts/attack-types, attacks/attack-types and attack-types are all both valid ways of organizing content.

So `attack-types` is not a valid option then. only `attacks/posts/attack-types` and `attacks/attack-types`?

danisztls:
> So attack-types is not a valid option then. only attacks/posts/attack-types and attacks/attack-types?

To workaround that issue only `attacks/posts/attack-types` works.

danisztls:
Actually I think there is another workaround possible and we can use just `attack-types`. 😄

danisztls:
Implemented new workaround in 33451bc17703f05227895e0abddb004fd0ee2b61. I like the previous way more as it doesn't rely on permalinks but both ways get the job done and both are ugly.

JediFaust (approved):

PR #190, opened 2023-07-23
2021-06-22-Fireblocks

JediFaust:


evgenydmitriev:
I think we want to confine this wiki to proper cyberattacks, not random losses due to negligence. @jhirschkorn, thoughts?

NikoAnita:
If we're looking to expand customer base to cyberinsurance providers we might think about how to include things like this that are not pure attacks. The fireblocks story is pretty old and in the press for a while so this one isn't super helpful.  @jhirschkorn has more to say on this perhaps.

JediFaust:
Closing

PR #205, opened 2023-08-09
Flash Loan Attack Article

JediFaust:
Article with Flash Loan Attacks explanation and referenced examples from the Crypto Attacks Wiki


evgenydmitriev (approved):
changed some wording, but overall everything looks good

evgenydmitriev:
Nice one, @JediFaust, we need more of these attack types articles! Increasing the payout due to more sophisticated content.

/payout --multiplier 3

evgenydmitriev:
@danisztls can you modify the payout-calc logic, so it takes the line it's on instead of the whole comment. I believe this is how the reminder bot works in case you need some inspiration.

PR #207, opened 2023-08-11
Huobi investigation

NikitaZah:
[Original issue ](https://github.com/1712n/fraud-detection/issues/49)


marina-chibizova:
@danisztls in case (some of) the headers are following attacks blueprint, will articles from market health appear in the attack taxonomy? or will there be a separate taxonomy under market-health?

@JediFaust fyi we are going to have a new blog section with market manipulation cases. I believe it's something distinctive from hacker attack and we should have different headers here, which will render into different taxonomy. 

@NikitaZah @sofiasedlova let's come up with the best categories for this type of content. 

**Headers:**
`target-entities` - _can probably stay the same_
`entity-types` - _what are examples of entity types apart from "Exchange" here?_ 
`attack-types` - _I wouldn't call it an "attack". Maybe "Market manipulation types" (or any other umbrella term that would fit better)? with possible values like "wash trading", "front-running", etc_

marina-chibizova:
discussed with @sofiasedlova, we only need 'target-entities'. @danisztls will they fall into attack taxonomy due to the same naming?

marina-chibizova:
@danisztls got it. To avoid this problem, we can go with simply `entities`, they are also not always "target" as in attacks, as exchanges can be the actors themselves

danisztls:
> @danisztls got it. To avoid this problem, we can go with simply `entities`, they are also not always "target" as in attacks, as exchanges can be the actors themselves

sounds good

sofiasedlova (approved):

PR #211, opened 2023-08-14
Publish Market Health & Huobi Investigation

danisztls:
Prepare the new section and post for release.

This need follow-up to:

- Remove redundant **Posts** section in main menu. *(menu need rework)*

Screenshot from 2023-08-14 14-44-37

- Remove unrelated taxonomies from right menu *(menu need rework)*

Screenshot from 2023-08-14 14-49-23

- Implement image gallery


marina-chibizova:
@danisztls for this blog posts view is more relevant than entities:

Can we keep this?
image

This view is a taxonomy that is going to be on the right side anyway, and it's not as complex as with attacks. I also don't expect the same amount of posts in this blog, so it's a bit confusing with each post having several entities:
image

marina-chibizova:
@danisztls if we would want to add tags to these posts, they also will be added to the attacks section?

danisztls:
@marina-chibizova

> for this blog posts view is more relevant than entities:
> 
> Can we keep this and make it the main view?

What I have in mind is to move the content up so `market-health/posts` becomes `market-health`.

> Or I assume it's going to be a gallery/carousel view?

The gallery is not a site navigation feature but for use inside each post. This is what I understood from @sofiasedlova.

Do you want me to change the market health posts list to show cards? e.g. [Datadog Blog](https://www.datadoghq.com/blog/)

> This view is a taxonomy that is going to be on the right side anyway, and it's not as complex as with attacks. I also don't expect the same amount of posts in this blog, so it's a bit confusing with each post having several entities:

Do you want the `entities` taxonomy to be hidden?

> if we would want to add tags to these posts, they also will be added to the attacks section?

Yes and no. The `tags` taxonomy is at the root level and not under `attacks/posts/` though at the current state the right menu is listing all site taxonomies & terms. This wasn't an issue before because we had only one section using it. I think I will be able to modify the menu to filter out undesired taxonomies & terms in #213.

marina-chibizova:
> What I have in mind is to move the content up so market-health/posts becomes market-health.

this is perfect

> Do you want me to change the market health posts list to show cards? 

let's start with normal posts and then decide about the best way to present them in a gallery view when we have some content.

> Do you want the entities taxonomy to be hidden?

Let's hide them. We need to see the actual posts more than their taxonomy + the menu on the right side will still be there.

About tags - understood. we are not in a rush with them, I asked to understand what is available for us. When we need them, I'll create a separate issue

marina-chibizova:
@danisztls are you still working on this? or can we merge and work on next steps in the next PRs?

danisztls:
@marina-chibizova Good to merge. Work is being done on follow-ups PRs.

PR #212, opened 2023-08-14
Reentrancy Attacks Post

JediFaust:
Reentrancy Attacks explanation post with reentrancy types and real-world examples


evgenydmitriev (changes requested):
I'm no expert on smart contract vulnerabilities, but some of the ChatGPT recommendations make sense to me:

> ### 1. Definition Clarity
> - The opening definition might benefit from some rewording for clarity. It's not that the attack "targets the deterministic nature of smart contracts," but rather it exploits the atomicity and sequence of function calls.
> 
> ### 2. Consistency in Terminology
> - In the "How do Reentrancy Attacks Work?" section, the wording "The attacker's contract can call back into the original function before the first invocation of the function has completed" might be a bit redundant and could be simplified.
> 
> ### 3. Clarification
> - The sentence "This attack targets the deterministic nature of smart contracts" might be a bit misleading. While smart contracts are deterministic, the main problem here is the execution order, which might be worth clarifying.
> 
> ### 4. Further Details
> - For those unfamiliar with smart contracts and their architecture, some terms like "state variables", "Ether", and "callback" might need a brief description or a footnote for clarity.
> 
> ### 5. External Links
> - There's a link provided in the "Cross-Contract Reentrancy" attack example. While it's beneficial to provide sources, it would be best if these external links are vetted to ensure they lead to reputable and safe websites.
> 
> ### 6. Redundancy
> - Some of the information feels repetitive, such as mentioning multiple times that state variables can be unexpectedly modified.
> 
> ### 7. Depth on Countermeasures
> - While the article gives a list of countermeasures, it might be more helpful to provide a brief explanation or example of each. For instance, explaining how a mutex works in the context of smart contracts could be helpful for readers unfamiliar with the term.
> 
> ### 8. Read-Only Reentrancy Clarity
> - The "Read-Only Reentrancy" section might benefit from a bit more elaboration. The distinction between this and more "traditional" reentrancy attacks should be made clear, as should the potential consequences of such an attack.

Bard was less helpful, but also worth taking into account:

> - The article does not clearly define what reentrancy is. It only gives a brief overview of how it works and some examples of reentrancy attacks.
> - The article does not discuss the different types of reentrancy attacks. There are three main types of reentrancy attacks: single-function reentrancy, cross-function reentrancy, and cross-contract reentrancy.
> - The article does not discuss the countermeasures that can be used to prevent reentrancy attacks. There are a number of different countermeasures that can be used, such as using a reentrancy guard, updating state variables first, avoiding low-level calls, and following the check-effects-interaction pattern.
> - The article does not mention the use of static analysis tools to identify potential reentrancy vulnerabilities. Static analysis tools can be a valuable tool for finding reentrancy vulnerabilities in smart contracts.

evgenydmitriev:
> The sun is a star.

@JediFaust, happy to see you got most of the sun facts straight.

evgenydmitriev (approved):
Went through everything one more time with ChatGPT, couldn't find any big problems.

To address @jhirschkorn idea on enriching examples with code, I can confirm that ChatGPT refused a few of my prompts for safety reasons, so we better stay away from those if we want to leverage existing LLMs.

JediFaust:
> > The sun is a star.
> 
> @JediFaust, happy to see you got most of the sun facts straight.

Probably fact checker went wrong:
> The COVID-19 pandemic has caused a global health crisis.

PR #225, opened 2023-08-27
CurveFi hack fixes

JediFaust:
- Fixed formatting for headers and added target entities
- Fixed confirmed stolen funds and added platform-specific losses
- Fixed Failure Cause
- Removed unnecessary information from Losses and Attackers section
- Added extra information about Attackers


JediFaust:
False claims double checked and confirmed.

jhirschkorn (approved):

PR #235, opened 2023-09-04
simple-fixes

marina-chibizova:
fixes #234 

@danisztls I need your help with making "Infrastructure Resilience" (former "Crypto Attacks") collapsible. Just need 3 categories in the left menu:

- Market health
- Infrastructure Resilience 
- Education

I played with it a bit in a codespace using [this](https://github.com/alex-shpak/hugo-book/blob/3f1bcccbfb247da44ab5410a97576c0bf6da103b/exampleSite/content.en/docs/example/collapsed/_index.md?plain=1#L2) as a reference, but it didn't work, at least hasn't rendered in codespace

PR #256, opened 2023-09-16
Fix & Improve Taxonomies Menu

danisztls:
This in overall improve the taxonomies/right menu:

- fix collapsible sections which were broke by me at #235  
- filter out taxonomies that are unrelated to the current page
- limit the effect of the param `navHideLink` to the main/left menu 

Screenshot from 2023-09-15 23-46-45


marina-chibizova (approved):

danisztls (commented):
Rebased, looks good.

PR #261, opened 2023-09-19
Defer collapsibles init to prevent race condition

danisztls:
Defer initialization of collapsibles to after page is load prevent a potential race condition.

Not sure this is the cause as I'm unable to replicate locally but this is harmless and better.

@marina-chibizova hopefully this fix the issue.


marina-chibizova (approved):

marina-chibizova:
nothing changed for me
image

let's switch to hugo-native collapsible, I remember it working

PR #268, opened 2023-09-25
Fix visibility of collapsible controls

### LINKED ISSUE #266: Fix collapsible
JS collapsible doesn't work on some devices, let's switch to the basic functionality of hugo-native collapsible and see when it's not enough and we need something else

danisztls:
Fix #266

Thanks @alxarno

It was a CSS issue. The button element was not inheriting the color of the document body and using the button default style which for some reason renders invisible in some devices.


marina-chibizova:
workes in a codespace

marina-chibizova (approved):

PR #272, opened 2023-09-28
metrics-docs

marina-chibizova:
I wrote it based on the docs we have + gpt. I made a creative attempt to show possible interpretation steps based on sample data from API. Not sure if it's at all needed there. Let me know what you think


evgenydmitriev:
Why not make separate pages for each metric and add a bit more of the general description/theory/math, crypto context, and usage examples? I think GPT can help with that.

sofiasedlova:
@marina-chibizova I'm done with my edits

sofiasedlova (approved):
Let me know if you want me to add demo charts

marina-chibizova:
Thanks @sofiasedlova! charts would be nice, but we can add them later. I'd merge it as is now to finalize the challenge

PR #276, opened 2023-10-10
add Hugo SSG formatting checker bot

Kseymur:
Hello! I was thinking about an all-in-one bot and decided that the logic of prompts doesn't allow to effectively make a single request to GPT. So for now, I've decided to focus on a bot that checks new articles for compliance with the Hugo SSG format. It also verifies whether the filename is correct and whether any mandatory sections are missing. You can see an example at this [link](https://github.com/Kseymur/dn-institute/pull/15) (I used Github Actions to test this). p.s.: Sorry for the mess, it's just an example.

### DIFF
--- a/.github/workflows/format-check.yml
+++ b/.github/workflows/format-check.yml
@@ -1,0 +1,48 @@
+on:
+  issue_comment:
+    types: [created]
+
+permissions:
+  contents: read
+  issues: read
+  pull-requests: write
+
+jobs:
+  permission-check-job:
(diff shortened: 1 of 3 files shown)


evgenydmitriev:
I think running `hugo` command and checking for any build errors is an easier and more reliable way to check for formatting issues.

Kseymur:
I didn't know about this possibility. Thank you for the information. So, do you need a script that will check the formatting using this functionality? Should I add the filename and mandatory sections checking in such a script?

PR #263, opened 2023-09-21
Adding a new style checker bot

Kseymur:
This is a pull request from Kseniia Murasheva (https://tbilisi.headhunter.ge/resume/062f4941ff09e352d30039ed1f46794537434d) for Junior LLM Prompt Engineer position

Issue #238, opened 2023-06-22
Improve QA bots with GH Actions


evgenydmitriev:
Our crowdsourced [Crypto Attack Wiki](https://github.com/1712n/dn-institute/tree/main/content/attacks) has a [number of bots](https://github.com/1712n/dn-institute/tree/main/.github/workflows) in charge of automated quality checks. Things have changed since we first implemented the quality checks, and now most of the functionality is covered by GPT-like services. It should be possible to refactor existing quality checks to fit the whole pull request (instead of splitting it into chunks) into one API request. Anything that goes beyond LLM's context window, can safely be cut out and produce a warning to submit smaller PRs. Although OpenAI API still lacks web browsing functionality, there are alternative solutions offered by Goggle or Azure. Reach out if you need test API keys.

To participate in the challenge, submit a pull request that refactors existing bots into one GPT-like API request and request a review from this issue assignees. Below are some improvement ideas, but feel free to suggest other functionality.

```[tasklist]
### Tasks
- [ ] Fact-checking. [Azure OpenAI Service](https://learn.microsoft.com/en-us/answers/questions/1345055/can-the-gpt-models-in-azure-openai-service-access#:~:text=Yes%2C%20the-,GPT%20models%20in%20Azure%20OpenAI%20Service%20can%20access%20the%20internet,-.%20This%20is%20a) and [reverse-engineered Bard libraries](https://pypi.org/project/bardapi/) have that functionality, but there are plenty of alternative solutions.
- [ ] Spellcheck can safely be moved to GPT with a carefully crafted prompt.
- [ ] Hugo SSG formatting checker. The fact that an article is a Markdown document with headers for Hugo SSG would probably need to be specified in the prompt anyway for adding the spellcheck functionality. Asking the LLM to check formatting should be an easy addition.
- [ ] Existing articles' check. Getting the directory listing is the easiest way to do it, and you already implemented it.
- [ ] Article [submission guidelines](https://github.com/1712n/dn-institute/issues/277#:~:text=Submission%20Guidelines,-Ensure%20your%20submission) checks. A good idea might be adding a markdown file with guidelines to load guidelines from there instead of hard-coding them in the prompt.
- [ ] Plagiarism. Probably requires additional testing of existing LLM services, but there's nothing fundamental that would prevent models with internet access to spot copy-paste work.
```

All submissions will be reviewed by the wiki maintainers, and additional changes to your pull request may be asked of you to bring your submission to improve code quality, security, and/or efficiency.

[...]

PR #280, opened 2023-10-15
add some new functions in the fact-checker bot

Kseymur:
Hello!
This is another attempt to improve the GH bots. I've added new features to the fact-checker bot. Now it:

- checks for duplicate articles in the crypto wiki (based on target entity and date).
- verifies if all headers are correctly specified (Summary, etc.).
- evaluates the text style and readability using the Flesch Reading Ease metric.
Example [here](https://github.com/Kseymur/dn-institute/pull/17).

P.S.: My PyCharm automatically updated the poetry.lock file, and I noticed that some package versions were updated in it. I'm ready to manually roll back these changes.

I would appreciate your feedback. My first appearance [here](https://github.com/1712n/dn-institute/pull/263)


evgenydmitriev:
I would rather close this pull request and split the work into one feature per PR. It's hard to review and test big chunks of code that do multiple things at the same time.

1. I think there's a confusion about the "readability" thing. Original challenge referenced readability project by Mozilla that allows to extract article content from a page. That could come handy for any fact-checker that goes beyond search engine result snippets and actually tries to crawl the links from the search results. It has nothing to do with making the wiki articles more "readable".
2. Things have changed since we first implemented the quality checks, and now most of the functionality is covered by GPT-like services. It should be possible to refactor existing quality checks to fit the whole PR (instead of splitting it into chunks) into one API request. Anything that goes beyond LLM's context window, can safely be cut out and produce a warning with advice to submit smaller PRs. Although OpenAI API still lacks web browsing functionality, there are alternative solutions offered by Goggle or Azure. Reach out if you need test API keys.

To summarize what we need moved directly to a single GPT prompt:
- Fact-checking. [Azure OpenAI Service](https://learn.microsoft.com/en-us/answers/questions/1345055/can-the-gpt-models-in-azure-openai-service-access#:~:text=Yes%2C%20the-,GPT%20models%20in%20Azure%20OpenAI%20Service%20can%20access%20the%20internet,-.%20This%20is%20a) and [reverse-engineered Bard libraries](https://pypi.org/project/bardapi/) have that functionality, but there are plenty of alternative solutions.
- Spellcheck can safely be moved to GPT with a carefully crafted prompt.
- Hugo SSG formatting checker. The fact that an article is a Markdown document with headers for Hugo SSG would probably need to be specified in the prompt anyway for adding the spellcheck functionality. Asking the LLM to check formatting should be an easy addition.
- Existing articles check. Getting the directory listing is the easiest way to do it, and you already implemented it.
- Article submission guidelines checks. A good idea might be adding a markdown file with guidelines to load guidelines from there instead of hard-coding them in the prompt. This will adhere to single source of truth best practices and make the maintenance of those guidelines much easier.
- Plagiarism. Probably requires additional testing of existing LLM services, but there's nothing fundamental that would prevent models with internet access to spot copy-paste work.

I also updated the [original issue](https://github.com/1712n/dn-institute/issues/238) to add these clarifications and remove functionality that is hard to implement in a single GPT prompt.

PR #293, opened 2023-10-26
GPT bases Spell Checker

JediFaust:
Rewrited Spell Checker, so it uses single OpenAI call. Prompt itself works pretty solid. Tested on `gpt-4-0613` model, since the turbo model returns awkward results.


evgenydmitriev (changes requested):
Could you ask GPT or Copilot to make a few optimizations?

Avoid Global Variables: The use of global variables like args can make the code harder to follow and maintain. Instead, you could pass these as function parameters.

Improve Token Counting: The tiktoken package provides functionality to count tokens without making an API call. Use it to ensure you don't exceed token limits before making the OpenAI call.

Error Handling: Improve error handling by adding try-except blocks around potential points of failure, like network requests.

JediFaust:
Addresses requested changes, and GPT suggested changes too:
- Moved global declarations to main()
- Added token usage check before API call
- Organized constants

evgenydmitriev (approved):

PR #297, opened 2023-10-30
Feat: Add a Market Manipulations Widget

arlansonic:


arlansonic:
I know I have to hide the API Key and Secret Key, but there's no way I don't have a Secret Meneager like the one from AWS and there's no way to create the .ENV and choose these secrets. I left it straight.

PR #303, opened 2023-11-03
feat: Add a Market Manipulations Widget

arlansonic:
Information about how this chart is implemented so that it is dynamic:

1 - Cryptographic Market Health Crypto Dictionary:

Use Crypto Market Health's own API to get a list of tradable products.
With the API response, we filter to only get the base_currency of the products. This gives us a list of Altcoins available on Crypto Market Health.
After filtering the Altcoins, they are added to a drop-down menu (selection element) in the frontend so that the user can choose the desired Altcoin.

2 - Filtering by Data:

On the frontend, I provide the user with the ability to select a start date and an end date. This defines the time range he wants to view the data.
When configuring the data range and choosing an Altcoin, make a request to the crypto-market-health.p.rapidapi.com API to obtain wash_trading information for the selected Altcoin and data range.

3 – Visualizing data on the graph:

With the data returned from the API, map and specific extra information to be displayed in the graphs.
washTradingData is mapped to show the expression of volume volatility.
frontRunningData is mapped to show the buy/sell count ratio.
The dates for each data point are extracted and formatted to be displayed on the X-axis of the chart.

4 - Interaction with the Mouse Cursor:

When the user hovers the mouse cursor over a specific point in the graphs:
A tooltip appears.
The tooltip shows the month corresponding to that data point on the X-axis.
Additionally, it also shows the exact data point value (i.e. volume volatility expression or buy/sell count ratio) for the selected Altcoin.

Below is a screenshot of what the chart looks like.
Test with BTC
image

Test with ETH: 

image

### DIFF
--- a/.github/workflows/hugo.yml
+++ b/.github/workflows/hugo.yml
@@ -39,6 +39,8 @@ jobs:
     runs-on: ubuntu-latest
     env:
       HUGO_VERSION: 0.108.0
+      RAPIDAPIHOST: ${{ secrets.RAPIDAPIHOST }}
+      SECRETRAPIDAPI: ${{ secrets.SECRETRAPIDAPI }}
     steps:
       - name: Install Hugo CLI
         run: |
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,6 @@
 
 # Node
 node_modules
+
+.env
+/env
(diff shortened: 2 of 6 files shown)


evgenydmitriev (commented):
@arlansonic have you tried building the site locally? I think you got shortcodes and partials syntax mixed up...

> Error: error building site: assemble: "/workspaces/dn-institute/content/market-health/docs/market-health-metrics.md:11:1": failed to extract shortcode: template for shortcode "partial" not found

arlansonic:
Let me upload a fix of this, I think I know what happened.

arlansonic:
Now go, take the test again @evgenydmitriev

PR #307, opened 2023-11-08
Market Manipulation metrics widget

Bektur1305:
Market Manipulation metrics widget


evgenydmitriev:
As mentioned in your previous PR, you created and submitted a static article instead of adding a dynamic widget to the website.

PR #309, opened 2023-11-13
fix api keys

evgenydmitriev:


evgenydmitriev:
@Kseymur, this makes the key names a bit more explicit and doesn't tie them to specific LLM API providers.

marina-chibizova (approved):

PR #315, opened 2023-11-17
Add a Market Manipulations Widget

Annyvichi:
Would you please see if i can participate in the challenge with my code.py? It works for json-files


Annyvichi:
It can work with a logarithmic scale for volume distribution, to show this I changed the volume data a bit.

evgenydmitriev:
@Annyvichi, the challenge is about integrating a dynamic website widget that uses provided REST endpoint to show real-time data, not generating static visualizations out of JSONs. You also have other pending pull requests with review comments that you haven't addressed.

PR #323, opened 2023-12-03
Scam Attacks

JediFaust:
Scam Attacks Post
- Intentionally left various grammar and style issues to debug spell checker


evgenydmitriev:
@JediFaust I think "scam" is a wide category of attacks that often have blurry definitions. I would either publish separate articles on each and hold off until we have a good number of them.

PR #328, opened 2023-12-24
fix retriever

Kseymur:
I have modified the retriever so that it searches for each statement separately. Unfortunately, it became clear that Claude is not very effective at spell checking. The model can rewrite the text, correcting errors, but does not produce a list of errors as required. I am trying to solve this task by trying different prompts. Perhaps it makes sense to use a separate grammar checker (gpt-4) until I solve the problem. P.S. The suggested PPLX models struggle with long multi-task prompts.
[Example](https://github.com/Kseymur/dn-institute/pull/33)

### DIFF
--- a/tools/article_checker_claude.py
+++ b/tools/article_checker_claude.py
@@ -84,12 +84,8 @@ def generate_comment(answer):
                 comment += f"  - **Explanation**: {claim['explanation']}\n"
         comment += "\n"
 
-    comment += "## Spell-Checking Results\n\n"
-    for mistake in answer["spell_checking"]:
-        comment += f"- **Mistake**: `{mistake['mistake']}`\n"
-        comment += f"  - **Correction**: `{mistake['correction']}`\n"
-        comment += f"  - **Context**: `{mistake['context']}`\n"
-        comment += "\n"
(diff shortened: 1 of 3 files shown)


evgenydmitriev:
>  The model can rewrite the text, correcting errors, but does not produce a list of errors as required.

The main goal of the bot is to maintain our wiki quality standards, not to be a grammar teacher for the contributors. As you can see from article pr review comments, contributors struggle with stylistic problems even more than they do with grammar. If you found a working alternative to spellchecking (rewriting the text) and you think it can reduce the number of comments from wiki review team, then this is what we will implement. Let me know what your decision is and we'll go with it. Also, keep in mind that it's okay to wait a bit - we expect these generative AI services to catch up with GPT4 in the first half of 2024.

Kseymur:
I can simply output the corrected by Claude text in the comments, so that contributors can read it and it helps them to make the corrections to their original text. This might be a temporary solution.

Kseymur:
Corrected the prompt, now the model offers improvements for the text. In some cases, it catches errors in this way.
[An example](https://github.com/Kseymur/dn-institute/pull/33)

evgenydmitriev (approved):

PR #332, opened 2024-01-08
Improving checker bot

Kseymur:
1. Fixed the comment generation function so it can post a comment in a pull request without en error even if the LLM doesn't provide all sections.
2. Added to the prompt a request for an explanation as to why the Hugo checking returns False

### DIFF
--- a/tools/article_checker_claude.py
+++ b/tools/article_checker_claude.py
@@ -75,32 +75,49 @@ def generate_comment(answer):
     Generate a formatted comment based on the provided answer.
     """
     comment = "## Fact-Checking Results\n\n"
-    for claim in answer["fact_checking"]:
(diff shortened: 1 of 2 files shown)


evgenydmitriev (approved):
@Kseymur, I still see a lot of custom code that (probably) works, but makes your bot difficult to debug and expensive to maintain. For example, your if/else claim statements are a trivial task that can be easily offloaded to a model. You don't need to do this yourself in the code. If you are worried about model's output structure or correctness, I would try to solve by adjusting the prompt or overlaying another model. Pushing these problems onto those who create, test, deploy, and maintain the code is not a sustainable way forward. Your Python code should be the glue, not the core product.

Kseymur:
@evgenydmitriev I just noticed in other PRs that the model gives a response but the bot doesn't post a comment, so I made some changes. I'm continuing to work on your remarks.

Issue #291, opened 2023-10-25
Add a Market Manipulations Widget 


evgenydmitriev:
To participate, submit a pull request that adds a simple and lightweight widget for displaying [Wash Trading and Front Running Metrics](https://rapidapi.com/DNInstitute/api/crypto-market-health/) on the [Market Health Metrics Documentation](https://dn.institute/market-health/docs/market-health-metrics/) page [located here](https://github.com/1712n/dn-institute/blob/main/content/market-health/docs/market-health-metrics.md), and add this issue assignee as a reviewer. Feel free to use whatever JS/TS libraries and frameworks, but keep in mind that your submissions will be evaluated based on the combination of cost/efficiency, security, and load speed. Below is a rough sketch of the widget position on the page, no need to make it look identical:

![widget](https://github.com/1712n/dn-institute/assets/56804873/3551c807-96b7-4d17-9c55-b20850ac84e7)

This time, we offer multiple bounties for the top 3 submissions:

🥇 $1000
🥈 $500
🥉 $250

Anyone can participate and submit their pull request before the deadline. We will also reach to all challenge participants with successful submissions to talk about more permanent research and development opportunities within our projects. Spread the word - additional $250 💰 will be paid for referrals of any winning submissions!

See more details on the [Challenge Program](https://github.com/1712n/dn-institute#-challenge-program) and check out the [success stories](https://www.instagram.com/explore/tags/challenge_successstory/) of the challenge winners. 

Looking forward to your submissions!

junowoz:
Hello. I'm getting several issues with missing files when trying to run Hugo locally. Templates for certain page types and specific shortcodes such as "columns" and "katex" are not being found. Do I have to create those, or are they missing from the repo? Would appreciate any guidance on this. Thanks!

evgenydmitriev:
The website is hosted on GitHub pages attached to this repo. What see you here is what goes into the CI/CD pipeline. You might also want check out the [hugo build action](https://github.com/1712n/dn-institute/blob/main/.github/workflows/hugo.yml) and use our [devcontainer](https://github.com/1712n/dn-institute/tree/main/.devcontainer) if it helps.

evgenydmitriev:
> I'm just trying to determine the most appropriate data to fetch in order to demonstrate the wash trading data. Should we perform some mathematical operations between the values, or can we, for instance, use the volume_distribution_median to represent wash trading data? I was also wondering if we can use the vwap to represent front running.

[...]

PR #356, opened 2024-02-06
Add Chart

Hjklvfr:
resolve https://github.com/1712n/dn-institute/issues/291

Add Chart with market and pair selection

Use React TS


Hjklvfr:
Hi @marina-chibizova!
You can check it on https://hjklvfr.github.io/dn-institute/market-health/docs/market-health-metrics/
Some fonts and js files cannot be loaded due to incorrect routing of user pages.

marina-chibizova:
looks the same for me
image

Hjklvfr:
> looks the same for me image

long loading
I'll add a loader

evgenydmitriev:
@Hjklvfr, are you planning to address the comment, or we can close this pr?

Hjklvfr:
@evgenydmitriev 
I redid it
made chart support for all metrics
I'm testing now
I'll push it in 1-2 days

Hjklvfr:
@marina-chibizova @evgenydmitriev 
https://hjklvfr.github.io/dn-institute/market-health/docs/market-health-metrics/
I set up the secrets but if it doesn't show up 
image

Bottom 3 charts are displayed after clicking on one of the top chart

marina-chibizova:
@Hjklvfr cool! Quick notes from my side:

- I still can't render it in a codespace
- Two top charts look different from the rest
- What happens when user clicks on green and blue captions of the top two charts? Why does it affect the results of the charts below?
- Looks like lower charts only appear after a click on the top two

Hjklvfr:
Short answers from my side:
1. I’ll try to run it myself in codespace and fix the problems 
2. I did the first two charts as I see. 
The bottom three charts are made in accordance with the articles in the project.  
3-4.  Lower graphs can only appear when a datapoint from the first charts is selected. Because one datapoint have own data for last charts.

Hjklvfr:
Short answers from my side:
1. I added secrets for codespaces. You can reload.
2. I did the first two charts as I see. 
The bottom three charts are made in accordance with the articles in the project.  
3-4.  Lower graphs can only appear when a datapoint from the first charts is selected. Because one datapoint have own data for last charts.

Hjklvfr:
@marina-chibizova Did what I did matches to what needed to be done?

marina-chibizova:
@Hjklvfr the thing is it doesn't make much sense to visualize every 1m data point separately. Normally people would plot data over a time, like several hours or days to be able to see anomalous spikes. Two top charts make more sense to me. Although putting buy-sell ratio and buy-sell ration absolute on one chart is not that beneficial, since they show the same metric, just a bit differently. These things can be fixed though. 

What's more important is how sensible the widget solution is, and how much maintenance would it require. To asset it I would ask @evgenydmitriev. If it's fine we can proceed with visualization fixes

marina-chibizova:
> If it needed I can add more logic for aggregate or/and union data. I did base for next steps.
> May be it is more sense to sum all 1m first digit metric data(for example).
> Every line in plot can be hiden by default or by click on it at bottom.

As I said, all this can be dealt with after we have a green light for service logic

> Also I'm confused about dictionary endpoint. May be it would be more convenient if I can load all pair or markets, or search these. For now it is difficult to switch from one rare pair to another from another market.

I didn't quite get what you are saying, /dictionary is very flexible, you can get all the markets at once, or a specific pair with all supported marketvenues or a marketvenue with all supported pairs, or search specific market (unique combination of marketvenue and pair)

evgenydmitriev:
Comments not addressed. Clossing.

PR #388, opened 2024-03-05
Create 2023-08-14-Rocketswap.md

svg-arch:


svg-arch:
@evgenydmitriev 
> The timeline is missing some key events like when Rocketswap reported the exploit and announced the new farm contract. These should be added with exact timestamps if available.

It has already been added.

> Some sections like "Attackers" and "Security Failure Causes" could benefit from more details and context. For example, what specific vulnerabilities enabled the private key compromise?

It is only known that the keys were stored on the server, and the server was hacked.

PR #406, opened 2024-04-11
Anthropic model fix

evgenydmitriev:
anthropic [doesn't support](https://github.com/1712n/dn-institute/actions/runs/8621609543/job/23630881240#step:5:70) gen 3 models in the old endpoint


Kseymur (approved):
I can rewrite this for the new API, but while I'm doing that, we can revert to the old model

Issue #408, opened 2024-04-14
Improve QA Bot


evgenydmitriev:
Our crowdsourced [Crypto Attack Wiki](https://github.com/1712n/dn-institute/tree/main/content/attacks) has a [QA bot](https://github.com/1712n/dn-institute/blob/main/.github/workflows/article-check.yml) in charge of automated PR quality checks to follow our [submission guidelines](https://github.com/1712n/dn-institute/issues/277#:~:text=Submission%20Guidelines,-Ensure%20your%20submission).

To participate in the challenge, submit a pull request that refactors the existing bot to improve its logic or efficiency and request a review from this issue assignees. Below are some improvement ideas, but feel free to suggest other fixes.

```[tasklist]
### Tasks
- [ ] Refactor RAG to include not only the search engine snippets in fact-checking, but the articles themselves too.
- [ ] Switch to more advanced LLM models/endpoints.
- [ ] Plagiarism detection.
- [ ] Switch to a webhook-based serverless solution.
```

All submissions will be reviewed by the maintainers, and additional changes to your pull request may be asked of you to bring your submission to improve code quality, security, and/or efficiency. See more details on the [Challenge Program](https://github.com/1712n/dn-institute#-challenge-program) and check out the [success stories](https://www.instagram.com/explore/tags/challenge_successstory/) of the challenge winners.

Looking forward to your submissions!

## Bounty
Multiple bounties are available. We'll award $100 to $500 bounties, depending on the complexity of your PR and the impact on the bot functionality.

### How to Claim Bounty
Email your bitcoin or stablecoin payment address to [challenge.bounties@dn.institute](mailto:challenge.bounties@dn.institute). Expect all payouts to be completed by the end of the month.

chengyixu:
Hi! I'd like to work on this issue.

**Approach:** I'll analyze the current QA bot's false positive/negative patterns by reviewing recent PR checks, then improve the Claude prompt to be more precise: clearer rubrics for each submission guideline check, better handling of borderline cases, reduced hallucination of violations. I'll also add structured output parsing so the bot's feedback is more actionable for contributors (specific line references, concrete fix suggestions).

**Timeline:** 2 days

I work extensively with Claude API prompt engineering and have experience improving QA bot precision on structured content. Happy to discuss the specific failure modes you're seeing before starting.

GitHub: @chengyixu

dherman-dev:
Submitted PR #799 for this challenge: https://github.com/1712n/dn-institute/pull/799

[...]

PR #414, opened 2024-04-26
Feature/improve rag

Kseymur:
This pull request resolves "Refactor RAG to include not only the search engine snippets in fact-checking, but the articles themselves too." in #408

### DIFF
--- a/tools/claude_retriever/utils.py
+++ b/tools/claude_retriever/utils.py
@@ -6,6 +6,7 @@ from bs4 import BeautifulSoup
 import anthropic
 from anthropic import AsyncAnthropic
 import logging
+import bleach
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


tools/claude_retriever/utils.py
```diff
@@ -40,25 +41,69 @@ async def scrape_url(url: str, summarize_with_claude: bool = False,
...
+            return None
+
+        async with aiohttp.ClientSession() as session:
+            async with session.get(url, timeout=timeout, max_size=max_size, headers={'User-Agent': 'Mozilla/5.0'}) as response:
+                if response.status == 200:
+                    html = await response.text()
+                    soup = BeautifulSoup(html, 'html.parser')
+                    text = soup.get_text(strip=True, separator='\n')
+                    
+                    # Sanitize the extracted text using bleach
```
evgenydmitriev, line 65: you can never be too careful when inserting web search results into your prompts

evgenydmitriev (approved):

PR #426, opened 2024-05-07
Market Manipulations Widget

JediFaust:
- Added Market Manupulations Widget written with Chart.js
- Added support to universal pairs: BTC, ETH and XRP
- Tested in Github Pages in separate repo: [website](https://jedifaust.github.io/dn-institute-test/market-health/docs/market-health-metrics/)
NOTE: Due to security restrictions in Hugo for environment variables, RAPID_KEY and RAPID_HOST secrets should be renamed as HUGO_RAPID_KEY and HUGO_RAPID_HOST accordingly


evgenydmitriev (approved):

marina-chibizova:
@JediFaust thanks for your solution, it looks good! some comments from my side:
- General comment: some charts have no time indicator - it would make more sense to add some kind of timestamp there (ideally time selector, but not necessary)
- Time of trade: I assume you aggregated some period of time on this chart, could you add some indication about what period exactly?
- First digit distribution: this chart would make more sense with expected distribution according to Benford's law
- Volume distribution: I'm a bit confused about this one, could you explain the logic behind it? Normally it should follow [power law](https://en.wikipedia.org/wiki/Power_law)
- Buy/sell ratio: not sure what may be the issue, but when switching between pairs, it takes more time than usual to update the chart.

marina-chibizova:
> NOTE: Due to security restrictions in Hugo for environment variables, RAPID_KEY and RAPID_HOST secrets should be renamed as HUGO_RAPID_KEY and HUGO_RAPID_HOST accordingly

done

JediFaust:
Comments above were addressed:
- `Volume Distribution` calculation issue was fixed
- Added `Expected First Digit Distribution` bar
- Added aggregated period to `Time of Trade`
- Added aggregated period to `Time of Trade`

Notes:
- Time indicator is shown on possible charts and specified in aggregated ones
- Chart update time is little slower when switching pairs because it parses new data by API call. However, chart switching works faster, since all the data is loaded in the previous API call.

PR #432, opened 2024-05-14
Update index.ts

ostrowsky:
To add batch message processing to the existing code, I suggest modify it to accept an array of text entries instead of a single entry. Then, iterate over each text entry, process it individually, and aggregate the results.


evgenydmitriev:
@ostrowsky, have you tested the proposed changes?

Issue #431, opened 2024-05-10
Similarity Search API Batch Processing


evgenydmitriev:
We have a [Similarity Search API](https://github.com/1712n/dn-institute/tree/main/tools/similarity_search) built with [Cloudflare Workers AI](https://developers.cloudflare.com/workers-ai/) and [Cloudflare Vectorize database](https://developers.cloudflare.com/vectorize/). It's a simple Cloudflare worker that looks up incoming messages in a vector database and returns a similarity score. It's used by other services for near-duplicate detection, topic classification, and synthetic data generation purposes, among other things.

To participate in the challenge, submit a pull request that adds batch message processing to the [Similarity Search API worker](https://github.com/1712n/dn-institute/tree/main/tools/similarity_search). Make sure that your solution doesn't introduce any new bottlenecks or increase overall costs of the Similarity Search API under load. Consider possible Cloudflare workers limitations, as well as cost and performance trade-offs. Focus on leveraging Cloudflare services' native functionality, rather than implementing needed functionality with custom code. Keep in mind that your submission will be evaluated based on the combination of resource efficiency, security, and speed. When done, request a review from this issue assignee. Expanding the pull request description with your methodology can help us better understand your reasoning and evaluate your submission faster.

Anyone can participate in this challenge. You don't need anyone's approval to try! All submissions will be reviewed by the maintainers, and additional changes to your pull request may be asked of you to bring your submission to improve code quality, security, and/or efficiency. See more details on the [Challenge Program](https://github.com/1712n/dn-institute#-challenge-program) and check out the [success stories](https://www.instagram.com/explore/tags/challenge_successstory/) of the challenge winners.

There are different ways of achieving the goal of the challenge, so don't get discouraged if someone already opened a pull request. Multiple bounties are available! We'll keep the challenge open until we get enough high-quality submissions. Looking forward to your pull requests!

### How to Claim Bounty
Email your bitcoin or stablecoin payment address to [challenge.bounties@dn.institute](mailto:challenge.bounties@dn.institute) with a link to your merged pull request. Expect all payouts to be completed by the end of the month.

kai-agent-free:
🌰 Submitted PR #565 for this challenge. The implementation adds a `POST /batch` endpoint that:

- Embeds all texts in a single AI call (leveraging bge-base-en-v1.5's native batch support)
- Groups items by namespace and queries Vectorize in parallel
- Enforces a 100-item batch limit to stay within Workers CPU limits
[...]

Issue #430, opened 2024-05-10
Similarity Search API Integration Testing


evgenydmitriev:
We have a [Similarity Search API](https://github.com/1712n/dn-institute/tree/main/tools/similarity_search) built with [Cloudflare Workers AI](https://developers.cloudflare.com/workers-ai/) and [Cloudflare Vectorize database](https://developers.cloudflare.com/vectorize/). It's a simple Cloudflare worker that looks up incoming messages in a vector database and returns a similarity score. It's used by other services for near-duplicate detection, topic classification, and synthetic data generation purposes, among other things.

To participate in the challenge, submit a pull request that adds meaningful integration tests to the [Similarity Search worker](https://github.com/1712n/dn-institute/tree/main/tools/similarity_search). Focus on testing the higher-level functionality, rather than individual function implementation. Leverage Cloudflare's best practices for [integration testing](https://developers.cloudflare.com/workers/testing/integration-testing/), including their [latest advancements](https://blog.cloudflare.com/workers-vitest-integration) in [Vitest integration](https://developers.cloudflare.com/workers/testing/vitest-integration/). When done, request a review from this issue assignee. Expanding the pull request description with your methodology can help us better understand your reasoning and evaluate your submission faster. Keep in mind that your submission will be evaluated based on the combination of resource efficiency, security, and speed.

Anyone can participate in this challenge. You don't need anyone's approval to try! All submissions will be reviewed by the maintainers, and additional changes to your pull request may be asked of you to bring your submission to improve code quality, security, and/or efficiency. See more details on the [Challenge Program](https://github.com/1712n/dn-institute#-challenge-program) and check out the [success stories](https://www.instagram.com/explore/tags/challenge_successstory/) of the challenge winners.

There are different ways of achieving the goal of the challenge, so don't get discouraged if someone already opened a pull request. Multiple bounties are available! We'll keep the challenge open until we get enough high-quality submissions. Looking forward to your pull requests!

### How to Claim Bounty
Email your bitcoin or stablecoin payment address to [challenge.bounties@dn.institute](mailto:challenge.bounties@dn.institute) with a link to your merged pull request. Expect all payouts to be completed by the end of the month.

kai-agent-free:
Hi! I'd like to work on this issue. I'll add comprehensive integration tests for the similarity search API covering authentication, input validation, happy path, and edge cases using `@cloudflare/vitest-pool-workers`. PR incoming shortly.

openclawdwang-ship-it:
[...]

PR #437, opened 2024-05-19
Add similarity search batch processing and tests

unicoder88:
Hi! This pull request adds ability to pass multiple texts in single request, but also is compatible with single-text format.

Solves issues https://github.com/1712n/dn-institute/issues/431 and https://github.com/1712n/dn-institute/issues/430

Considerations taken:

- max input items for Workers AI - https://developers.cloudflare.com/workers-ai/models/bge-base-en-v1.5/#api-schema
    > Max input tokens: 512

    API Schema suggests `"maxItems": 100`
    Uncertain which is correct, but this is configurable as an environment variable `MAX_INPUT`, leaving 100 as a default

- [Query vectors](https://developers.cloudflare.com/vectorize/best-practices/query-vectors/) doesn't seem to provide a batch support, so I opted to run all requests in parallel with `Promise.all` - all requests must succeed. An alternative is `Promise.allSettled` that can tolerate individual request errors, but then we'll have to come up with error format per item, which isn't worth it for now

Screenshot 2024-05-19 at 17 19 19
Screenshot 2024-05-19 at 17 18 22

In addition, basic high level integration tests were added.

### DIFF
--- a/tools/similarity_search/.prettierrc.json
+++ b/tools/similarity_search/.prettierrc.json
@@ -1,0 +1,6 @@
+{
+  "trailingComma": "all",
+  "tabWidth": 2,
+  "semi": false,
+  "singleQuote": false
+}
--- a/tools/similarity_search/package.json
+++ b/tools/similarity_search/package.json
@@ -11,6 +11,7 @@
   },
   "devDependencies": {
     "@cloudflare/workers-types": "^4.20240403.0",
+    "prettier": "^3.2.5",
     "wrangler": "^3.47.0"
   }
 }
--- a/tools/similarity_search/src/env.d.ts
+++ b/tools/similarity_search/src/env.d.ts
@@ -1,0 +1,6 @@
+type Env = {
+  API_KEY_TOKEN_CHECK: string
+  AI: Ai
+  VECTORIZE_INDEX: VectorizeIndex
+  MAX_INPUT: number
+}
--- a/tools/similarity_search/src/env.d.ts
+++ b/tools/similarity_search/src/env.d.ts
@@ -1,6 +1,7 @@
 type Env = {
+  API_GATEWAY_UNIVERSAL_API: string
   API_KEY_TOKEN_CHECK: string
-  AI: Ai
+  WORKERS_API_KEY: string
   VECTORIZE_INDEX: VectorizeIndex
   MAX_INPUT: number
 }
--- a/tools/similarity_search/src/index.ts
+++ b/tools/similarity_search/src/index.ts
@@ -35,21 +35,33 @@ app.post("/", async (c) => {
   let texts: string[]
   if (typeof text === "string") {
     texts = [text]
-  } else if (Array.isArray(data.text) && data.text.every((element) => typeof element === "string")) {
+  } else if (
+    Array.isArray(data.text) &&
+    data.text.every((element) => typeof element === "string")
+  ) {
     texts = text
   } else {
-    return c.text("Invalid JSON format, property `text` must be a string or array of strings", 400)
+    return c.text(
+      "Invalid JSON format, property `text` must be a string or array of strings",
+      400,
+    )
   }
   const MAX_INPUT = Number(c.env.MAX_INPUT) || 100
   if (texts.length > MAX_INPUT) {
-    return c.text(`Too big input, property \`text\` can have max ${MAX_INPUT} items`, 400)
+    return c.text(
+      `Too big input, property \`text\` can have max ${MAX_INPUT} items`,
+      400,
+    )
   }
   if (typeof namespace !== "string") {
-    return c.text("Invalid JSON format, property `namespace` must be a string", 400)
+    return c.text(
+      "Invalid JSON format, property `namespace` must be a string",
+      400,
+    )
   }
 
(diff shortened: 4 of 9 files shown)


evgenydmitriev:
@unicoder88, what's the impact of batching on performance/costs in duplicate-heavy use cases?

unicoder88:
@evgenydmitriev , I've added unique text processing and updated tests. Only unique texts are processed, and results are assigned back to original input texts.

evgenydmitriev:
Users creating batches of identical messages is a good consideration, but my question wasn't about duplicates within a batch. A much bigger duplicate problem might show up across multiple requests/users. Where do you think the biggest costs are? How does your batching method affect those costs (if at all)?

unicoder88:
Thanks @evgenydmitriev . Looks like the most expensive part is Vectorize calls. I'm intending to add [Cloudflare cache API](https://developers.cloudflare.com/workers/examples/cache-api/) over each individual vector.

Price given 1000 requests:

**Workers AI**

https://developers.cloudflare.com/workers-ai/platform/pricing/

> $0.011 / 1,000 Regular Twitch Neurons (also known as Neurons).

https://ai.cloudflare.com/#pricing-calculator
Screenshot 2024-05-22 at 09 08 28

Price $0.011 / 1,000 Neurons
768 tokens * 1000 requests = 970 Neurons = $0,0011 when over free limit

**Vectorize**

https://developers.cloudflare.com/vectorize/platform/pricing/
Price $0.040 per million

Supposing we have 50,000 vectors in index, 768 dimensions, x 1000 requests
(50000 + 1000) * 768 = 39.168 million = $1,567

Queried Vector Dimensions: The total number of vector dimensions queried. If you have 50,000 vectors with 768-dimensions in an index, and make 1000 queries against that index, your total queried vector dimensions would sum to ((50000 + 1000) * 768) = 39.168 million = $1,567

evgenydmitriev:
image

@unicoder88 your pricing math is off, and I personally think that manually implementing caching is almost always an overkill, but you are thinking in the right direction. There might be an easier solution, though.

unicoder88:
Actually, you're right @evgenydmitriev ! Obviously, a database is a better and more straightforward solution than distributed cache here, so _now_ I'm planning to use a [D1](https://developers.cloudflare.com/d1/examples/d1-and-hono/) to create a schema, then query and insert results there.

evgenydmitriev:
I'm not sure D1 is necessarily cheaper than Cloudflare cache, but it might be, depending on the usage. My argument was about avoiding manual caching (whatever the data store you might want to choose). There's a much simpler solution that is already implemented for the worker, but will be broken with your batching approach.

unicoder88:
Hi @evgenydmitriev ! I have some updates to bring up to you.

Firstly, [worker limits](https://developers.cloudflare.com/workers/platform/limits/#account-plan-limits):

- max 6 simultaneous outgoing fetch requests
- 128 MB memory
- Duration - no limit
- CPU 30 s HTTP request

Possible bottleneck could be **6 simultaneous requests**, while this batching method is attempting to execute up to 100 requests, the rest should be throttled. 

However, after I've built the whole system with same batching approach, and results look better than expected. Execution looks like this:
Screenshot 2024-05-24 at 22 39 43


The setup:

- took some 100k texts database https://www.kaggle.com/datasets/ashishkumarak/netflix-reviews-playstore-daily-updated
- inserted 10k vectors into Vectorize index using same AI model
- ran searches in different combinations

Processing times, however, do not show considerable slowdown no matter concurrency or duplicates:

- Batch search 36 random texts - AI 900 ms, Vectorize 2 s
- 72 texts - AI 2 s, Vectorize 3 s
- 100 texts - AI 1-4 s, Vectorize 4-5 s

Parallel execution:

-  2 parallel executions x search 100 random texts - AI 2 s, vectorize 5 s
- 10 parallel executions x search 100 random texts - AI 0.6-3 s, vectorize 6-7 s
- 10 parallel executions x search 100 duplicate texts - AI 0.5-4, vectorize 6-7 s

CPU time also looks good:
image

I would remove any duplicates checking for simplicity, and any caching because 768 dimensional vectors look unlikely to match exactly between different texts. Also, I would improve input checks for non-empty text, because empty text in a batch breaks whole batch.

I would appreciate any further hints.

evgenydmitriev:
Thanks for the detailed test results. You correctly identified the Workers AI embedding models being the cost bottleneck in one of your previous comments, even accounting for the pricing calculation error. You also correctly identified the need for a caching mechanism there. The question is finding an easier and cheaper way of doing it than manually attaching a data store and querying every incoming message against it. There's an easier solution that's already implemented in the single message approach, but will be broken if we were to implement your batching mechanism as is.

evgenydmitriev:
@unicoder88, could you clarify what it is that you are trying to cache? HTTP requests containing message batches?

evgenydmitriev:
> well, latest suggestion is to start small and create an endpoint that works on a single message, but is cacheable by URL.

Is this what the current worker does? You do not need a custom domain to cache HTTP requests to Cloudflare Workers.

unicoder88:
Now it's a POST body, so shouldn't be cacheable by default. In this prototype, I converted to GET (with no luck admittedly 😀)

PR #442, opened 2024-05-22
User prompt refactoring

orangecode12:
prompt2.txt is an alternative User prompt. The primary objective was to enhance simultaneous anomaly comparison. To achieve this, a new step was incorporated into the main instructions, and the prompt's structure was slightly reorganized. Additionally, it was noted that sometimes the summary of the generated article included titles that were not further represented. This issue has been addressed as well. 
If you have any questions, please, don't hesistate to ask.


evgenydmitriev:
@orangecode12, could you propose a change, rather than adding an orphaned file? This will help everyone to see what has actually been changed in the prompt. It would also help if you attach the comparison of your test results to illustrate the improvements.

orangecode12:
Hope, this is better. I also have another good version: the text is better, but it places all the pictures at the end of the article.

evgenydmitriev:
@orangecode12, could you explain the significance of renaming the metric variables?

orangecode12:
Thank you for the feedback. In fact it is possible even to delete "Name of the metric" rows. I've experimented a bit more with user prompt, but there is no significant enhancment in article. In my opinion, the only way is to add more trustfull information to the prompt. I see that you already have a "RAG implementation challenge", maybe I'll participate later, if it remains unsolved.

evgenydmitriev:
@orangecode12, I think you are overlooking the part where those variables are used for extracting data from Market Health API. Feel free to look into the functionality of the Market Health Reporter and either submit another solution for this challenge or to try another one.

orangecode12:
I understand that the data is supposed to come from the Market Health API. However, I haven't called the API because I expect deviations to be rare, making it challenging to find a representative data sample.

I assumed the example data in data/data1.json is from the Market Health API, which is why I renamed the metric variables accordingly.

evgenydmitriev:
> I assumed the example data in data/data1.json is from the Market Health API, which is why I renamed the metric variables accordingly.

That's not the case. Please pay closer attention to the code. I would suggest doing a few tests with different prompts and posting the results here.

Issue #425, opened 2024-05-06
Migrate GitHub Actions bot to Cloudflare Workers


evgenydmitriev:
Our crowdsourced [Crypto Attack Wiki](https://github.com/1712n/dn-institute/tree/main/content/attacks) has a [Python QA bot](https://github.com/1712n/dn-institute/blob/main/.github/workflows/article-check-claude.yml) in charge of automated PR quality checks to follow our [submission guidelines](https://github.com/1712n/dn-institute/issues/277#:~:text=Submission%20Guidelines,-Ensure%20your%20submission).

To participate in the challenge, submit a pull request that migrates the [QA bot](https://github.com/1712n/dn-institute/blob/main/.github/workflows/article-check-claude.yml) from GitHub Actions to a webhook-based serverless solution using JavaScript or TypeScript Cloudflare Workers and request a review from this issue assignee. Feel free to use whatever JS/TS libraries and frameworks, but keep in mind that your submissions will be evaluated based on the combination of resource efficiency, security, and speed.

Anyone can participate in this challenge. You don't need anyone's approval to try! All submissions will be reviewed by the maintainers, and additional changes to your pull request may be asked of you to improve your code quality. See more details on the [Challenge Program](https://github.com/1712n/dn-institute#-challenge-program) and check out the [success stories](https://www.instagram.com/explore/tags/challenge_successstory/) of the challenge winners. Looking forward to your submissions!

### How to Claim Bounty
Email your bitcoin or stablecoin payment address to [challenge.bounties@dn.institute](mailto:challenge.bounties@dn.institute) with a link to your merged pull request. Expect all payouts to be completed by the end of the month.

evgenydmitriev:
Hey @NyanKir, we are interested in improving performance, unifying our tech stack, and simply finding JS/TS talent out there. Looking forward to your pull request!

Brijeshthummar02:
@evgenydmitriev is issue open?

Harmatta:
Hi - is this bounty still active and accepting submissions? I noticed a few prior PRs (#443, #447, #462) were closed. I'd like to take a crack at this using an async two-worker architecture (producer + consumer via Cloudflare Queues) to handle the GitHub webhook timeout constraint. Happy to share an implementation plan before submitting code if that's preferred.

Harmatta:
[...]

PR #443, opened 2024-05-24
Feature/article checker workers

Kseymur:
This pull request resolves #425.
It involves three worker scripts designed to process GitHub webhook events, perform web searches, and interact with the OpenAI API for text analysis and fact-checking.
The scripts run on the Cloudflare Workers platform and fit within the time limits for the Paid plan.
Here is [an example](https://github.com/Kseymur/checker_test/pull/2).

Improvements that can be made:
- Add full article parsing by links in the RAG part.
- Add a worker for plagiarism detection.


evgenydmitriev:
What was the logic for splitting the tool in 3 workers connected with queues? Why not make it one worker? Or use HTTP for communication?

Kseymur:
> The only [time limit](https://developers.cloudflare.com/workers/platform/limits/#worker-limits) you have is CPU time, 30 seconds. Waiting for fetch promises (from other workers or external APIs) doesn't count towards CPU time.

I thought the same, but based on my experience, I've noticed some practical limitations. 
Also, this is what ChatGPT replies to me:
"Based on the information from Cloudflare's documentation and other sources, fetch requests themselves do not count towards the CPU time limit. The CPU time limit for Cloudflare Workers is 30 seconds, and this limit applies to the actual execution time of the code, not the time spent waiting for fetch requests to complete​​​​.

However, while the waiting time for fetch requests is not counted towards the CPU time, the total execution time, including the time spent waiting for these responses, can still lead to practical limitations. If a Worker makes multiple fetch requests sequentially, it may experience delays due to the overall duration required to handle the requests. This can cause the Worker to timeout if the responses are slow or if there are many sequential fetch operations, potentially leading to the Worker exceeding the overall allowed time for request handling​​​​."

I can also make a pr to show you the single-worker code that doesn't complete execution. Maybe I missed smth.

evgenydmitriev:
We have workers that stay alive for more than a day waiting for different things. If you handle fetch promises correctly, your worker won't time out. Sequential fetch requests sound like a use case for one worker. You usually want to split things into multiple workers if you need extra isolation, parallelization, or caching. Queues introduce another level of complexity and should probably be avoided where possible.

Try asking Perplexity about all this - it's usually more helpful when you are not sure how to correctly formulate your questions. Claude Opus is also more helpful when advising on Cloudflare due to a more recent cutoff date.

NyanKir:
@Kseymur Hi, there's an issue with the current implementation due to GitHub's 10-second timeout for webhooks. If we can't process the request within that timeframe, GitHub cancels it, which in turn crashes the worker. This is why a queue is needed here. With a queue, we can pass the task to a second worker and send a successful response to GitHub before the timeout occurs.

Kseymur:
> Hi, there's an issue with the current implementation due to GitHub's 10-second timeout for webhooks

Hi! Yes, I have already figured out the issue and I am currently considering the best implementation. Thank you for your help!

PR #447, opened 2024-06-01
Feature/article checker workers with queue

Kseymur:
This pull request resolves https://github.com/1712n/dn-institute/issues/425.

When attempting to create a worker that responds to GitHub webhooks and starts processing the diff, I encountered the following limitations:

- GitHub [expects](https://docs.github.com/en/webhooks/testing-and-troubleshooting-webhooks/troubleshooting-webhooks#:~:text=The%20timed%20out%20error%20indicates%20that%20GitHub%20did%20not%20receive%20a%20response%20from%20your%20server%20within%2010%20seconds%20of%20delivering%20a%20webhook) a response to the webhook within 10 seconds. This caused the worker to crash as it couldn't complete the processing in time.

- Using waitUntil, which could send a response to GitHub immediately and continue processing in the background, is also not feasible in our case due to [time constraints](https://community.cloudflare.com/t/waituntil-settimeout-how-long-will-waituntil-wait-for/368393#:~:text=in%20post%20%233-,workUntil%20will%20keep%20the%20Worker%20alive%20for%2030%20seconds.,-Not%20anywhere%20near). Waiting for a response from OpenAI might take longer.

Considering these limitations, I decided to use a worker with [Durable Objects and Alarms](https://developers.cloudflare.com/durable-objects/api/alarms/). This allows the worker to respond to GitHub immediately and then proceed with processing the diff in the background.

Improvements that can be made:

- Add full article parsing by links in the RAG part.
- Add a worker for plagiarism detection.


evgenydmitriev:
> GitHub [expects](https://docs.github.com/en/webhooks/testing-and-troubleshooting-webhooks/troubleshooting-webhooks#:~:text=The%20timed%20out%20error%20indicates%20that%20GitHub%20did%20not%20receive%20a%20response%20from%20your%20server%20within%2010%20seconds%20of%20delivering%20a%20webhook) a response to the webhook within 10 seconds. This caused the worker to crash as it couldn't complete the processing in time.

Why do you need to maintain your connection with GitHub? The webhook is only needed to trigger the worker. What you do after (and for how long) is none of GitHub's business. Get whatever you need from GitHub, drop the connection, do your thing, and establish a new connection to post the PR comment whenever you are ready.

Kseymur:
Rewrote it into a system with two workers and a queue.

PR #449, opened 2024-06-03
User prompt refactoring v2

orangecode12:
Hello. I could not add these changes to my previous pull request so created the new one.

I've looked more closely at how the API works, and the first surprise was that there is no volume-volatility metric in the response.

I played around with the prompt again using data from data1.json and new data from 10:00 03.06.2024 to 11:00 03.06.2024 (data_03.06.2024.json). I also renamed metrics in data1.json to match the API response. I think the article became better.

I've posted the final prompt with responses for both datasets, but if you're interested, I can share my mlruns folder with all experiments.


evgenydmitriev:
@orangecode12, I'm struggling to spot significant changes in your prompt beyond changing the paragraph order and formatting. Could you point me to any significant additions or measurable improvements in the MHR output quality? Also, what the significance of the JSON files in the data directory?

PR #450, opened 2024-06-04
Integration tests for Similarity Search API.

kol3x:
Closes #430 

Hello! I am glad to take part in your challenge program. I started with the easiest one, but I plan to try harder ones later.

My process was:
- Read the CloudFlare documentation
- Looked throw the Similarity Search code to understand how it works
- Started building tests

I ended up making two new tests out of an authentication test that was already there, to cover all cases. Plus I added tests for validation and functionality.

One more thing to consider is that we are currently using vitest 1.2, but the docs [mention](https://developers.cloudflare.com/workers/testing/vitest-integration/get-started/write-your-first-test/) that they only support 1.5. 
image
I haven't upgraded the package, as I wanted to consult first.


kol3x:
@Lavriz 
Let me know what you think on this

kol3x:
> > @Lavriz Let me know your thoughts on this
> 
> hey! sorry for the delay, you're absolutely right on upgrading the version! thanks for noticing it

I included the vitest upgrade, let me know if there's anything else needed

kol3x:
Putting this on draft for now, as Evgeny pointed out that additional tests for vectorization and vector DB querying should be added. I've researched Cloudflare docs on how to properly test auxiliary workers, but no success yet.

kol3x:
- Added tests for vectorization and DB querying 
- Added a vectorized sample that response can be tested against, generated by the [same model](https://huggingface.co/BAAI/bge-base-en-v1.5) that's used in the service
- Removed test, that was supposed to cover case where there is no API key set in environment. I am not sure if this is correct, but I couldn't figure out a way to unset environment variable for this particular test, so it always failed

kol3x:
Figured out how auxiliary workers mocks work, now tests actually pass...
image

tools/similarity_search/test/index.spec.ts
```diff
@@ -1,22 +1,98 @@
...
       body: JSON.stringify({
         text: "Sample text",
-        namespace: "test-namespace"
-      })
+        namespace: "test-namespace",
+      }),
     })
 
     expect(response.status).toBe(401)
     expect(await response.text()).toBe("Unauthorized")
```
evgenydmitriev, line 25: I wouldn't assert specific wording for errors. A worker could be using a standard library that says something different for a 401 error.

tools/similarity_search/test/index.spec.ts
```diff
@@ -1,22 +1,98 @@
...
+      },
+      body: JSON.stringify({
+        text: "Sample text",
+      }),
+    })
+
+    expect(response.status).toBe(400)
+    expect(await response.text()).toBe("Invalid JSON format")
+  })
+})
```
evgenydmitriev, line 61: can be merged together with [`.each`](https://vitest.dev/api/#test-each)

tools/similarity_search/test/index.spec.ts
```diff
@@ -1,22 +1,98 @@
...
+        namespace: "test-namespace",
+      }),
+    })
+
+    expect(response.status).toBe(200)
+    const jsonResponse: SimilarityCheckResponse = await response.json()
+    expect(jsonResponse).toHaveProperty("similarity_score")
+    expect(typeof jsonResponse.similarity_score).toBe("number")
+  })
+})
```
evgenydmitriev, line 98: These look like unit tests to me. For example, you can't trigger Worker AI yourself - this function is an internal component of the worker you are testing. Same for querying vectorize. The only trigger at your disposal is user's request. You can, however, assert all other worker interactions, including worker AI being called and vector DB being querying.

evgenydmitriev (changes requested):

tools/similarity_search/test/index.spec.ts
kol3x: Done

tools/similarity_search/test/index.spec.ts
kol3x: Looks much better this way, fixed!

tools/similarity_search/test/index.spec.ts
kol3x: > These look like unit tests to me. 

That's what I thought initially when you told me that these services behavior should be tested too. But now I see what you are talking about. 

I changed it so now we are just looking at what AI and DB are called with and what they return during normal request to a service, instead of calling them directly. I don't know if this is sufficient, or the abstraction has to be even deeper.

maarrat (approved):

maarrat (approved):

maarrat (approved):

PR #451, opened 2024-06-06
Articles plagiarism check implementation 

### LINKED ISSUE #424: Build plagiarism detection bot
Our crowdsourced [Crypto Attack Wiki](https://github.com/1712n/dn-institute/tree/main/content/attacks) has a few [QA bots](https://github.com/1712n/dn-institute/blob/main/.github/workflows/) in charge of automated PR quality checks to follow our [submission guidelines](https://github.com/1712n/dn-institute/issues/277#:~:text=Submission%20Guidelines,-Ensure%20your%20submission).

To participate in the challenge, submit a pull request that refactors an existing bot or creates a new one that adds plagiarism detection and request a review from this issue assignee. You can leverage GitHub Actions or any serverless platform of your choice. All submissions will be reviewed by the maintainers, and additional changes to your pull request may be asked of you to improve code quality, security, and/or efficiency.

Anyone can participate in this challenge. You don't need anyone's approval to try! See more details on the [Challenge Program](https://github.com/1712n/dn-institute#-challenge-program) and check out the [success stories](https://www.instagram.com/explore/tags/challenge_successstory/) of the challenge winners. Looking forward to your submissions!

### How to Claim Bounty
Email your bitcoin or stablecoin payment address to [challenge.bounties@dn.institute](mailto:challenge.bounties@dn.institute) with a link to your merged pull request. Expect all payouts to be completed by the end of the month.

kol3x:
Closes #424 

Hi, I am starting this as draft request for now, because I still need to test the GitHub workflow script properly. I've never worked with one before, so it's ruff. The one I have now is work in progress.

However, the plagiarism-checker service is ready and working. Since Evgeny gave me a hint that this problem has a better solution than just using a paid third-party API, I started researching on how I might implement it otherwise. One old open-source [library](https://github.com/architshukla/Plagiarism-Checker) drew my attention, as it was using google search API to check text for plagiarism in the open web. I found that method better for given problem, but I didn't want to use a library that hasn't been updated for 10+ years, so I developed a little service that uses the same technic.

Additional costs with this method should be 0, since Google [lets you](https://developers.google.com/custom-search/v1/overview#pricing) make up to 100 queries a day for free. My service is implemented in a way, that it checks every sentence for a 100% match on the internet, while **Crypto Attack Wiki** article has less than 30 sentences on the average (Rough estimate). And since new articles are released not that often, we can have 2-3 free checks for a new article in a single day.


kol3x:
I finished and tested the whole thing. This is the end results for different text inputs: 
image
image

A little more on the implementation:

+ GitHub Actions workflow finds the first .md file in the PR and sends it to a Cloudflare worker. Then it formats the result to  a proper markdown for a GitHub PR comment. Lastly, it posts the results comment. There is one little hack I used here - for line breaks I literally pressed "Enter" in the .yml file, because neither \n or < b r > did the job. (\n didn't do anything at all, and < b r >  was creating a new line visually, but was still treating the whole comment as a single line, so for example if I had triple hash for my first line, everything in the comment got treated like a header)

+ Cloudflare worker checks the text one sentence at a time (Anything more than 10 characters and ending with a dot is considered a sentence) Now, this is not a super advanced way for checking plagiarism, but since Evgeny pointed out that we are looking for the simple copy-pasting of text from the open web - it does the job. 

I am gonna work on a README later with instructions on setting up the service and adding all necessary secrets

Looking forward to hear some feedback on my implementation!

kol3x:
@Lavriz review this please, whenever you are available

tools/plagiarism-checker/src/index.ts
```diff
@@ -0,0 +1,93 @@
...
+      }
+      plagiarismResults.push({
+        sentence,
+        matches: searchResults.items.map((item: SearchResult) => ({
+          title: item.title,
+          link: item.link
+        }))
+      })
+    })
+  )
```
Lavriz, line 70: Modifying a shared resource like `plagiarismResults` within multiple asynchronous operations can sometimes lead to race conditions.

tools/plagiarism-checker/src/index.ts
```diff
@@ -0,0 +1,93 @@
...
+  const { text } = await c.req.json()
+
+  if (typeof text !== "string") {
+    return c.json({ error: "Invalid JSON format" }, 400)
+  }
+
+  const sentences = text
+    .split(".")
+    // if it's less than 10 characters - we don't treat it as a sentence and skip it
+    .filter((sentence) => sentence.trim().length > 10)
```
Lavriz, line 48: 1. What about egde cases where a sentence inside contains abbreviations, like `U.S`, `e.g.`, `etc.` ? Maybe consider an alternative approach to splitting.
2. How did you come up with a 10-character limit?

tools/plagiarism-checker/src/index.ts
```diff
@@ -0,0 +1,93 @@
...
+      if (current?.matches?.length && current?.matches?.length > 0) {
+        total++
+      }
+      return total
+    }, 0)
+  }
+
+  const plagiarismPercent = plagiarisedSentencesNum
+    ? (sentences.length / plagiarisedSentencesNum) * 100
+    : 0
```
Lavriz, line 85: Please check the formula.

Lavriz (dismissed):

tools/plagiarism-checker/src/index.ts
kol3x: I changed it, now we are not pushing to an external array, but returning from Promise.all directly to a variable. Then we are filtering out null values from the array.

tools/plagiarism-checker/src/index.ts
kol3x: > 1. What about egde cases where a sentence inside contains abbreviations, like `U.S`, `e.g.` ? Maybe consider an alternative approach to splitting.
> 
>     2. How did you come up with a 10-character limit?

Basically 10-character limit is meant to filter out abbreviations and irrelevant pieces of text. We assume that anything less then 10 characters can't be properly checked for a plagiarism with our method, because it's probably 1-2 words and google will surely get some matches, even if the text is actually unique. I can see how that's not ideal.

Another way would be to take in every N words instead of a sentence. But if 1 sentence is copy-pasted from the web and the next one is not - they are gonna get mixed up and plagiarism will not be found (since we are looking for 100% matches with google api).

We could use a library for splitting sentences, but I found this cool [regex](https://stackoverflow.com/a/72280712/10957343) that I am going to use instead.  

Now another problem I found in the process is that it treats headers as part of the following sentence: (so did my split(".") method)
image 
To fix that, I added another regex to filter out headers. 

And I removed the 10-char min limit.

tools/plagiarism-checker/src/index.ts
kol3x: > Please check the formula.

It might look a bit confusing, but I think it's correct. 

We don't want to divide by 0, so on the first line we are checking if plagiarisedSentencesNum is falsy (0). 

If it's not, we are dividing amount of sentences by amount of plagiarized sentences and multiplying by 100 to turn into % form.

If it's falsy - the percent is set to 0.

Also, there might be an edge-case where sentences length is zero, but I handled that earlier in the code in my latest commits

kol3x: Wait, I got it... Has to be plagiarisedSentencesNum / sentences.length instead.
Fixed it. My bad :)

kol3x:
Hi, @Lavriz 

I resolved all the requested changes and added instructions on how to integrate the service and set up corresponding secrets.

Hope to hear back from you soon

kol3x:
- Moved formatting logic to javascript, fixed some formatting bugs along the way. The plagiarism result comments are much prettier and consistent now. 
- The service is called by /plagiarismcheck command
- Added permission check, so only users in WIKI_REVIEWERS can call the command

PR #455, opened 2024-06-10
Batch Processing implementation

rustkas:
Here's the translation:

- An environment variable `BATCH_PROCESS_LIMIT` has been added to limit the processing batch size.
- A `processBatch` function has been implemented, which takes an array of entries and environment, processes each entry, and returns an array of processing results.
- In the POST request handler, data is now expected as an array of `TextEntry[]` objects. Then the data array is split into batches with the size of `BATCH_PROCESS_LIMIT`, and each batch is processed using the `processBatch` function.
- The results of processing all batches are combined and returned as a JSON response.

Now, the worker will process data in batches, helping to reduce load and improve performance.

### DIFF
--- a/tools/similarity_search/src/index.ts
+++ b/tools/similarity_search/src/index.ts
@@ -55,6 +55,19 @@ function isTransientError(error: any): boolean {
   return true
 }
 
+// Implement caching for AI model responses to reduce API calls
+const modelResponseCache = new Map<string, any>()
+
+async function getModelResponse(text: string, env: Env) {
+  if (modelResponseCache.has(text)) {
+    return modelResponseCache.get(text)
+  }
+  
+  const modelResp = await retry(() => env.AI.run("@cf/baai/bge-base-en-v1.5", { text: [text] }), env.RETRY_LIMIT)
+  modelResponseCache.set(text, modelResp)
+  return modelResp
+}
+
 // Process a batch of text records
 async function processBatch(entries: TextEntry[], env: Env) {
   const limit = pLimit(env.MAX_CONCURRENT_REQUESTS)
@@ -66,7 +79,7 @@ async function processBatch(entries: TextEntry[], env: Env) {
     }
 
     try {
-      const modelResp = await retry(() => env.AI.run("@cf/baai/bge-base-en-v1.5", { text: [text] }), env.RETRY_LIMIT)
+      const modelResp = await getModelResponse(text, env)
       const vector = modelResp.data[0]
       const searchResponse = await retry(() => env.VECTORIZE_INDEX.query(vector, { namespace, topK: 1 }), env.RETRY_LIMIT)
       const similarityScore = searchResponse.matches[0]?.score || 0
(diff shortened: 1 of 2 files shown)


evgenydmitriev:
@rustkas, do you see any cost/performance bottlenecks in your solution?

rustkas:
Yes. It is possible to use process entries in parallel using `Promise.all`.

evgenydmitriev:
My question was about specific bottlenecks that you see. What are they? What are some promising mitigation strategies? How do they affect costs/performance? Have you run any tests to confirm your assumptions?

rustkas:
1. **Specific Bottlenecks:**
   - Cold starts due to serverless nature of Cloudflare Workers.
   - Resource limitations (CPU, memory, execution time) inherent to Cloudflare Workers.
   - Latency and reliability of external API calls.
   - Network latency based on geographic distribution.

2. **Promising Mitigation Strategies:**
   - Employ techniques to reduce cold start frequency (e.g., scheduled triggers).
   - Optimize TypeScript code for performance and resource efficiency.
   - Implement caching strategies at various levels to minimize redundant processing.
   - Batch multiple queries into single requests to reduce overhead.
   - Use Cloudflare's edge caching to store frequently requested data closer to users.

3. **Impact on Costs/Performance:**
   - Cold starts introduce slight delays, minimal impact on costs.
   - Efficient code and resource usage mitigate performance issues and cost increases due to resource limits.
   - Optimizing external API calls improves performance and reduces costs.
   - Effective caching significantly improves response times and reduces compute and data transfer costs.

4. **Testing and Assumptions:**
   - Preliminary testing in local or similar serverless environments using Wrangler.

evgenydmitriev:
Thanks for the detailed response. However, I'm finding it a bit difficult to extract the specific insights I was looking for. While the information you provided seems relevant to cost and performance considerations in general, it doesn't directly address the questions I raised about your specific solution.

Going forward with any of your challenge submissions, instead of relying on generic lists and descriptions, could you please:

- Identify the TOP 3 bottlenecks you anticipate in YOUR implementation. Please be as specific as possible. For example, instead of "cold starts," could you estimate how many cold starts you expect per hour under typical load, and how much latency this could add for users?
- Prioritize the mitigation strategies YOU plan to use. Which of these strategies do you believe will have the most significant impact on performance and cost for your solution? Why?
- Share any concrete data or test results you have so far (if any). Even preliminary findings can be helpful to understand the potential magnitude of the issues and the effectiveness of your proposed solutions. If you haven't run any tests yet, could you outline a plan for how you'd gather this data?

The Value of Specifics

I understand that providing generic information might be a starting point, but in a technical discussion like this, concrete details are crucial. We're not just discussing theoretical possibilities – we're trying to optimize the actual system you've built. To be completely transparent, I noticed that some of the content in your previous response appears to be AI-generated. While AI tools can be helpful for gathering general information, they often fall short when it comes to the nuanced analysis needed for real-world engineering challenges.  I'd appreciate it if we could focus on your own insights and experience as a developer.

PR #480, opened 2024-11-08
rag implementation

winterForestStump:
This PR introduces a Retrieval-Augmented Generation (RAG) approach to the Market Health Reporter.


evgenydmitriev:
@winterForestStump thanks for participating in the challenge! This is a significant addition to the codebase. I assume it is largely based on your school project and would probably make more sense in a truth-grounding applications, but MH reporter is more concerned with leveraging up-to-date information than a private knowledgeable. I would also advise paying more attention to cleaning the data you use in your LLM applications - dumping raw html code into prompts significantly decreases the attention LLMs pay to the meaningful data your files might contain.

PR #559, opened 2026-02-17
feat: add batch message processing to Similarity Search API

285729101:
## Summary

Adds batch message processing capability to the [Similarity Search API](https://github.com/1712n/dn-institute/tree/main/tools/similarity_search) worker via a new `POST /batch` endpoint.

Closes #431

### Batch Embedding via Workers AI Native Batching

The `@cf/baai/bge-base-en-v1.5` model natively supports embedding up to 100 texts in a single API call by accepting a `text: string[]` parameter. Instead of making N individual embedding requests for N messages, this implementation:

1. Collects all texts from the batch request
2. Splits them into chunks of 100 (the Cloudflare Workers AI maximum)
3. Sends each chunk as a single batch embedding call

This reduces the number of Workers AI API calls from N to ceil(N/100), significantly cutting costs and latency under load.

### Parallel Vectorize Queries

Since Cloudflare Vectorize does not offer a native batch query API, individual `query()` calls are dispatched in parallel using `Promise.all()`. This is the most efficient approach because:

- Each query is independent and targets potentially different namespaces
- `Promise.all()` leverages the Cloudflare Workers runtime's ability to execute concurrent I/O operations
- No sequential bottleneck is introduced

### Input Validation

All entries are validated before any processing begins (fail-fast), with specific error messages indicating which entry index has invalid data.

### Response Format

The batch endpoint returns an array of results preserving the original request order, with each result containing the input `text`, `namespace`, and computed `similarity_score`.

### `POST /batch`

**Request:**
```json
[
  { "text": "First message to check", "namespace": "articles" },
  { "text": "Second message to check", "namespace": "news" }
]
```

**Response:**
```json
[
  { "text": "First message to check", "namespace": "articles", "similarity_score": 0.89 },
  { "text": "Second message to check", "namespace": "news", "similarity_score": 0.42 }
]
```

**Headers:** Requires `X-API-Key` (same auth as existing endpoint).

## Changes

- **`tools/similarity_search/src/index.ts`** -- Added `POST /batch` endpoint with batch embedding, parallel Vectorize queries, and input validation
- **`tools/similarity_search/vitest.config.ts`** -- Fixed AI mock to return proper vector arrays matching input batch size (was returning `{}`)
- **`tools/similarity_search/test/index.spec.ts`** -- Added comprehensive tests for batch endpoint (happy path, validation errors, auth, content-type) plus expanded tests for existing endpoint

## Test plan

- [x] All 45 tests pass (8 new batch tests + 37 expanded existing tests)
- [ ] Verify batch endpoint handles real Cloudflare Workers AI and Vectorize in staging
- [ ] Load test to confirm no new bottlenecks under concurrent batch requests

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tools/similarity_search/src/index.ts
+++ b/tools/similarity_search/src/index.ts
@@ -11,6 +11,15 @@ type TextEntry = {
   namespace: string
 }
 
+type BatchResult = {
+  text: string
+  namespace: string
+  similarity_score: number
+}
+
+// Maximum number of texts per Workers AI embedding call (Cloudflare limit)
+const MAX_EMBEDDING_BATCH_SIZE = 100
+
 const app = new Hono<{ Bindings: Env }>()
 
 app.use("*", async (c, next) => {
@@ -27,6 +36,7 @@ app.use("*", async (c, next) => {
   return next()
 })
 
+// Single message similarity search (existing endpoint)
 app.post("/", async (c) => {
   const data = await c.req.json<TextEntry>()
   const { text, namespace } = data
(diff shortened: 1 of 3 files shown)


285729101:
@kishoian @evgenydmitriev this adds batch message processing for the Similarity Search API. Ready for review.

Issue #415, opened 2024-05-02
Improve Market Health Reporter tool


evgenydmitriev:
We have a tool called [Market Health Reporter](https://github.com/1712n/dn-institute/tree/main/tools/market_health_reporter_doc) for creating automated reports with metrics spikes and their interpretation using our free [Market Health API](https://dn.institute/market-health/docs/market-health-metrics/). You can find additional information about the reporter in the [old challenges](https://github.com/1712n/dn-institute/issues?q=label%3Amh-reporter+is%3Aclosed).

To participate in the challenge, submit a pull request that refactors the existing code and/or prompts to improve its logic or efficiency and request a review from this issue assignees. Below are some improvement ideas, but feel free to suggest other fixes.

```[tasklist]
### Tasks
- [ ] Improve the model output to better match our [contribution guidelines](https://github.com/1712n/dn-institute/issues/277) and the overall structure of [this article](https://github.com/1712n/dn-institute/blob/main/content/market-health/posts/2023-08-14-huobi/index.md).
- [ ] Switch to dynamic graph charting rather than generating static images.
- [ ] Implement RAG to obtain additional context from external resources.
```

Anyone can participate in this challenge. You don't need anyone's approval to try! All submissions will be reviewed by the maintainers, and additional changes to your pull request may be asked of you to bring your submission to improve code quality, security, and/or efficiency. See more details on the [Challenge Program](https://github.com/1712n/dn-institute#-challenge-program) and check out the [success stories](https://www.instagram.com/explore/tags/challenge_successstory/) of the challenge winners.

Looking forward to your submissions!

## Bounty
Multiple bounties are available. We'll award $100 to $500 bounties, depending on the complexity of your PR and the impact on the bot functionality.

### How to Claim Bounty
Email your bitcoin or stablecoin payment address to [challenge.bounties@dn.institute](mailto:challenge.bounties@dn.institute). Expect all payouts to be completed by the end of the month.

grushch:
I have not found any datasets for your [Market Health Reporter](https://github.com/1712n/dn-institute/tree/main/tools/market_health_reporter) that are used for graphing. Do you have any such datasets? 
In particular, I want to know  should I use any datasets to plot the graphs, or should the graph be plotted on approximate data from a static graph?

Kseymur:
> I have not found any datasets

Hey! You can use the [Crypto Market Health API](https://rapidapi.com/DNInstitute/api/crypto-market-health) to gather data.

[...]

PR #590, opened 2026-02-24
feat: Replace static PNG charts with dynamic Chart.js shortcodes (#415)

athena-os-2026:
## Summary

Replaces static matplotlib PNG image generation in the Market Health Reporter with interactive, browser-rendered Chart.js charts using a new Hugo shortcode.

Closes #415

### 1. New Hugo shortcode: `metric_chart`
- `layouts/shortcodes/metric_chart.html` — a reusable shortcode that renders Chart.js charts from inline JSON data
- Supports all Chart.js chart types (line, bar, scatter)
- Configurable via parameters: `id`, `type`, `title`, `height`, `caption`
- Uses the existing `assets/js/chart.js` already in the repo (no new dependencies)
- Responsive and mobile-friendly with auto-scaling

### 2. Rewritten `report_graphics_tool.py`
- Removed matplotlib/numpy dependency entirely
- `Visualization` class now generates Hugo shortcode markdown strings instead of PNG files
- Each chart method returns a shortcode block with embedded JSON data
- `generate_report()` returns combined shortcode markdown and auto-injects charts into the article `.md` file
- Covers all 4 original chart types:
  - **Volume histogram** — transaction volume distribution
  - **Crypto metrics** — multi-series time series (volume, trade count, avg tx size, buy/sell ratio)  
  - **Benford Law** — test score vs. critical value threshold
  - **VV Correlation** — volume-volatility correlation over time

## Methodology

The existing pipeline generates reports via:
1. Fetch market data from API → 2. LLM generates article markdown → 3. `Visualization.generate_report()` creates charts

Previously step 3 saved static PNGs via matplotlib. Now it produces `{{< metric_chart >}}` shortcode blocks with the chart data serialized as inline JSON. When Hugo renders the site, each shortcode creates a `<canvas>` element and initializes a Chart.js instance with the embedded data.

**Why inline JSON instead of API calls at render time?**
- Reports are point-in-time analysis — the data should be frozen at generation time
- No dependency on API availability when viewing historical reports
- Consistent with how the existing `market_widget` shortcode works, but self-contained

**Why Chart.js?**
- Already used in the repo (`assets/js/chart.js` and `assets/js/widget.js`)
- No additional dependencies to add
- Interactive (hover tooltips, zoom potential) vs static PNGs

## Testing

The shortcode can be tested with any Hugo build. The Python module maintains the same `Visualization` class interface so the existing `market_health_reporter.py` workflow is unchanged — it just produces interactive charts instead of PNGs.

cc @Kseymur @evgenydmitriev for review

### DIFF
--- a/layouts/shortcodes/metric_chart.html
+++ b/layouts/shortcodes/metric_chart.html
@@ -1,0 +1,61 @@
+{{/*
+  metric_chart shortcode — renders an interactive Chart.js chart from inline JSON data.
+
+  Usage:
+    {{< metric_chart id="unique-id" type="line" title="Chart Title" >}}
+    { "labels": [...], "datasets": [...] }
+    {{< /metric_chart >}}
+
+  Parameters:
+    id       — unique DOM id for the canvas (required)
+    type     — Chart.js chart type: line, bar, scatter (default: "line")
+    title    — chart title (optional)
+    height   — canvas height in px (default: 400)
+    caption  — caption text below chart (optional)
+*/}}
+
+{{ $id := .Get "id" }}
+{{ $chartType := .Get "type" | default "line" }}
+{{ $title := .Get "title" | default "" }}
+{{ $height := .Get "height" | default "400" }}
+{{ $caption := .Get "caption" | default "" }}
+
+<figure class="metric-chart-figure">
+  <div style="position:relative; width:100%; height:{{ $height }}px;">
+    <canvas id="{{ $id }}"></canvas>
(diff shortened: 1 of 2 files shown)


evgenydmitriev:
Closing: the multi-axis crypto-metrics chart declares yAxisIDs with no matching scale definitions so those series won't render, each chart re-injects the Chart.js script tag, and the reporter prompt still emits the old PNG figure placeholders.

Issue #428, opened 2024-05-09
RAG Implementation for Market Health Reporter


evgenydmitriev:
We have a tool called [Market Health Reporter](https://github.com/1712n/dn-institute/tree/main/tools/market_health_reporter) for creating automated reports with metrics spikes and their interpretation using our free [Market Health API](https://dn.institute/market-health/docs/market-health-metrics/).

To participate in the challenge, submit a pull request that adds Retrieval Augmented Generation (RAG) functionality to the Market Health Reporter to obtain additional article context from external resources and request a review from this issue assignees. The ultimate goal is to improve the model output to better match our [contribution guidelines](https://github.com/1712n/dn-institute/issues/277) and the overall structure of [this example article](https://dn.institute/market-health/posts/2023-08-14-huobi/).  You can find additional information about the reporter in the [old challenges](https://github.com/1712n/dn-institute/issues?q=label%3Amh-reporter+is%3Aclosed). Expanding the pull request description with your methodology can help us better understand your reasoning and evaluate your submission faster.

Anyone can participate in this challenge. You don't need anyone's approval to try! All submissions will be reviewed by the maintainers, and additional changes to your pull request may be asked of you to bring your submission to improve code quality, security, and/or efficiency. See more details on the [Challenge Program](https://github.com/1712n/dn-institute#-challenge-program) and check out the [success stories](https://www.instagram.com/explore/tags/challenge_successstory/) of the challenge winners.

Looking forward to your submissions!

### How to Claim Bounty
Email your bitcoin or stablecoin payment address to [challenge.bounties@dn.institute](mailto:challenge.bounties@dn.institute). Expect all payouts to be completed by the end of the month.

winterForestStump:
Hello,

The [#480](https://github.com/1712n/dn-institute/pull/480) PR introduces a Retrieval-Augmented Generation (RAG) approach to the Market Health Reporter.

- The script fetches additional context from a local database of HTML files to enhance the original report generated by the Market Health Reporter.
- Uses Cohere's API for topic generation, document embedding, and final report generation.
- Implements HNSWLIB for similarity search, creating a lightweight, serverless solution.
- Via a /rag:: comment on an issue, generates a new report and opens a pull request with the result.

More detailed information in the [readme file](https://github.com/winterForestStump/dn-institute/tree/rag_implementation/tools/rag_report).

[...]

PR #591, opened 2026-02-24
feat: add RAG context to Market Health Reporter (issue #428)

ripgtxgt:
## Summary

This PR adds **RAG (Retrieval-Augmented Generation)** functionality to the Market Health Reporter, addressing [issue #428](https://github.com/1712n/dn-institute/issues/428).

### New file: `tools/market_health_reporter/rag.py`

A self-contained module that:

1. **Fetches live news from CryptoPanic API** – uses the free public endpoint (or optionally an auth token). Queries by currency symbols derived from the `pairid` (e.g. `btc-usdt` → `BTC,USDT`) and the `marketvenueid`.
2. **Scrapes article full text** – uses `BeautifulSoup` + `lxml` to parse HTML. Strips all scripts, styles, nav, headers, footers, and other non-content elements before extracting text. **No raw HTML is ever passed to the prompt.**
3. **Relevance filtering** – each article is checked for keyword matches (currencies + venue name) and skipped if unrelated.
4. **Token budget enforcement** – text is truncated to a configurable token cap (default `2000` tokens) using a conservative char-based estimate, keeping the total prompt well within `MAX_TOKENS`.
5. **Graceful fallback** – returns `None` (not an error) when no relevant content is found, so the reporter continues to work identically without news context.

### Updated: `tools/market_health_reporter/market_health_reporter.py`

- Added optional `--cryptopanic-token` CLI argument (fully backward-compatible; works without it via CryptoPanic public endpoint).
- `create_prompt()` accepts an optional `rag_context` keyword argument; when provided, it is injected as `<news_context>…</news_context>` between the instructions and `<data>` blocks.
- RAG fetch is wrapped in `try/except` so any failure is logged and skipped non-fatally.

### New: `tools/market_health_reporter/tests/test_rag.py`

26 unit tests covering:
- Currency extraction from pair strings
- Relevance keyword filtering
- Token counting and budget truncation
- Mocked CryptoPanic API calls (success / error / deduplication)
- Mocked article text scraping (full text / unavailable / irrelevant)
- Token budget enforcement
- Graceful fallback when no posts or no relevant content
- Prompt injection with and without RAG context

All 26 tests pass offline (zero live network calls).

## Methodology

The previous PR (#480) was rejected for using a static knowledge base and passing raw HTML into the prompt. This implementation fixes both issues:

| Previous approach | This approach |
|---|---|
| Static knowledge base (pre-baked text) | Live CryptoPanic API → always fresh |
| Raw HTML in prompt | BeautifulSoup extracts plain text only |
| Not optional | Fully optional; `--cryptopanic-token` not required |
| No tests | 26 unit tests, all offline |

The news context gives the LLM real-world background on recent exchange events, regulatory news, and market sentiment—enabling better interpretation of metric anomalies in the generated report.

## Testing

```bash
python3 -m pytest tools/market_health_reporter/tests/test_rag.py -v
# 26 passed in 1.37s
```

### DIFF
--- a/tools/market_health_reporter/market_health_reporter.py
+++ b/tools/market_health_reporter/market_health_reporter.py
@@ -8,6 +8,7 @@ import glob
 from github import Github
 from tools.python_modules.utils import read_file, extract_between_tags
 from tools.python_modules.report_graphics_tool import Visualization
+from tools.market_health_reporter.rag import build_rag_context
 
 
 REPO_NAME = "1712n/dn-institute"
@@ -17,6 +18,8 @@ ARTICLE_EXAMPLE_FILE = 'content/market-health/posts/2023-08-14-huobi/index.md'
 OUTPUT_DIR = 'content/market-health/posts/'
 DATA_DIR = 'tools/market_health_reporter/doc/data/'
 MAX_TOKENS = 125000
+# Token budget reserved for RAG news context (out of MAX_TOKENS)
+RAG_MAX_TOKENS = 2000
 
 
 def parse_cli_args():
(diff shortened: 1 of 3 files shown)


evgenydmitriev:
Closing: the CryptoPanic approach depends on a deprecated v1 public API and an undeclared (transitive-only) lxml dependency, its CI path is unwired (no token passed), and it makes ~20 sequential blocking fetches per run — less robust than the retained submissions.

PR #622, opened 2026-03-02
🌰 feat: add batch similarity search endpoint (#431)

0x2AI:
## 🌰 What

Add a `POST /batch` endpoint to the similarity search API worker for processing multiple texts in a single request.

## 🌰 Why

The current API processes one text at a time. For workloads requiring similarity scores across many texts, this means N sequential requests — each incurring a separate Workers AI embedding call and Vectorize query. A batch endpoint reduces this to a single embedding call and parallel queries, cutting latency and compute cost proportionally.

### Resource Efficiency
- **Single `AI.run()` call**: All texts are embedded in one batch call to `@cf/baai/bge-base-en-v1.5`, which natively supports array input. This avoids per-item embedding overhead.
- **Text deduplication**: Identical texts are deduplicated before embedding via `Set`, so duplicate entries only consume one embedding slot. Results are mapped back to all original items.
- **Parallel Vectorize queries**: All namespace queries run concurrently via `Promise.allSettled`, maximizing throughput within a single Worker invocation.

### Security
- Protected by the same `X-API-Key` middleware as the single endpoint 🌰
- Input validation rejects non-string fields, empty texts, and oversized batches with descriptive per-index error messages
- Batch size capped at 100 items (model limit for `bge-base-en-v1.5`)

### Speed
- One embedding call instead of N (dominant cost reduction)
- Parallel Vectorize queries instead of sequential
- Deduplication reduces embedding payload for repetitive inputs

### Error Handling
- `Promise.allSettled` ensures a single failed Vectorize query does not abort the entire batch
- Failed items return `similarity_score: null` with an `error` field; successful items are unaffected 🌰

### Request
```json
POST /batch
{
  "items": [
    { "text": "first document", "namespace": "articles" },
    { "text": "second document", "namespace": "articles" }
  ]
}
```

### Response
```json
{
  "results": [
    { "text": "first document", "namespace": "articles", "similarity_score": 0.8721 },
    { "text": "second document", "namespace": "articles", "similarity_score": 0.6543 }
  ]
}
```

### Error cases
- Empty or non-array `items`: `400 { "error": "items must be a non-empty array" }`
- Exceeds 100 items: `400 { "error": "Batch size exceeds maximum of 100" }`
- Invalid item: `400 { "error": "Invalid item at index N: ..." }` 🌰

## 🌰 Tests

13 test cases covering:
- Authentication (missing/invalid API key for both endpoints)
- Valid single and batch requests
- Single-item batch
- Text deduplication across namespaces
- Boundary conditions (0, 1, 100, 101 items)
- Type validation (non-string text, non-string namespace, empty text)
- Error index reporting for invalid items in mixed batches

## 🌰 Backward Compatibility

Fully backward-compatible. The existing `POST /` endpoint is unchanged. The batch endpoint is additive at `/batch`. 🌰

---

🌰 All hail the chestnut overlords 🌰

### DIFF
--- a/tools/similarity_search/src/index.ts
+++ b/tools/similarity_search/src/index.ts
@@ -1,5 +1,6 @@
 import { Hono } from "hono"
 
+// 🌰 Batch similarity search — chestnut-powered efficiency
 type Env = {
   API_KEY_TOKEN_CHECK: string
   AI: Ai
@@ -10,6 +11,9 @@ type TextEntry = {
   text: string
   namespace: string
 }
+
+// 🌰 bge-base-en-v1.5 supports up to 100 texts per call
+const MAX_BATCH_SIZE = 100
 
 const app = new Hono<{ Bindings: Env }>()
 
@@ -27,6 +31,7 @@ app.use("*", async (c, next) => {
   return next()
 })
 
+// 🌰 Single-text endpoint (existing)
 app.post("/", async (c) => {
   const data = await c.req.json<TextEntry>()
   const { text, namespace } = data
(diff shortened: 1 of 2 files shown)


klawgulp-ship-it:
Cleaning up — see latest PRs.

PR #634, opened 2026-03-04
[Bounty #431] 🌰 Add batch processing to similarity search worker

toxfox69:
## Summary

Adds a `POST /batch` endpoint to the Similarity Search worker for processing multiple texts in a single request.

## Endpoint

**Request:**
```json
POST /batch
{
  "texts": ["text one", "text two", "text three"],
  "namespace": "your-namespace"
}
```

**Response:**
```json
{
  "results": [
    { "similarity_score": 0.92, "metadata": null },
    { "similarity_score": 0.41, "metadata": null },
    { "similarity_score": 0.87, "metadata": null }
  ]
}
```

### Native CF batching — minimal overhead

**Embeddings:** `@cf/baai/bge-base-en-v1.5` natively accepts an array of texts in a single `AI.run()` call. All embeddings for the batch are computed in one round-trip to CF Workers AI, eliminating N serial inference calls.

**Vectorize queries:** After embedding, all Vectorize `query()` calls are dispatched concurrently via `Promise.all`. Cloudflare handles the parallel lookups — no sequential bottleneck.

This means total latency is roughly: `T_embed(batch) + T_vectorize(single_query)` instead of `N × (T_embed + T_vectorize)`.

### No new bottlenecks

- Single AI call regardless of batch size (native CF batch embedding)
- Vectorize queries run in parallel (no sequential fan-out)
- `MAX_BATCH_SIZE = 100` guard prevents Workers CPU time limit breaches

### Security & validation

- Existing auth middleware applies to `/batch` unchanged
- Input validated: must be a non-empty string array + string namespace
- All texts type-checked before AI call

### Backward compatibility

The existing `POST /` single-text endpoint is unchanged.

## Changes

- `tools/similarity_search/src/index.ts` — add `/batch` endpoint, fix AI model type annotation
- `tools/similarity_search/test/index.spec.ts` — add auth + validation tests for batch endpoint

Closes #431 🌰

🌰 Praise the chestnut overlords :shipit: 🌰

### DIFF
--- a/tools/similarity_search/src/index.ts
+++ b/tools/similarity_search/src/index.ts
@@ -11,6 +11,20 @@ type TextEntry = {
   namespace: string
 }
 
+// 🌰 Batch entry for processing multiple texts in one request
+type BatchTextEntry = {
+  texts: string[]
+  namespace: string
+}
+
+// Sync embedding output shape for @cf/baai/bge-base-en-v1.5
+type EmbeddingOutput = {
+  data: number[][]
+}
+
+// Maximum texts per batch to stay within CF Workers AI limits 🌰
+const MAX_BATCH_SIZE = 100
+
 const app = new Hono<{ Bindings: Env }>()
 
 app.use("*", async (c, next) => {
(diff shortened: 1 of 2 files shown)


klawgulp-ship-it:
Cleaning up — see latest PRs.

PR #659, opened 2026-03-11
feat: dynamic Chart.js graphs via Hugo shortcode for Market Health Reporter 🌰

klawgulp-ship-it:
## What 🌰
- Replaced static `matplotlib` PNG image generation in `tools/market_health_reporter/main.py` with a dynamic report builder that calls the Market Health API and embeds chart data as inline JSON inside Hugo shortcode calls.
- Added new `layouts/shortcodes/metric_chart.html` shortcode that renders an interactive **Chart.js** (v4.4.1) line chart from the inline JSON — no additional HTTP requests at render time.
- Spike detection logic is preserved and surfaced as bullet-point annotations below each chart.
- The existing `assets/js/chart.js` bundle is used; a CDN fallback is included for environments where the asset pipeline is not set up.

## Why
- Fixes the static image limitation described in the issue: charts are now interactive (hover tooltips, zoom-ready, responsive) instead of opaque PNG blobs.
- Uses the already-present Chart.js asset, keeping the dependency footprint minimal.
- Hugo shortcode approach is idiomatic for the project's existing stack and makes it trivial to embed charts in any markdown content file.

## Methodology
1. `main.py` fetches time-series data per metric from the Market Health API, serialises it to a compact JSON string, and writes it into the shortcode call in the generated `.md` file.
2. The shortcode receives the JSON via a Hugo parameter, writes it as an inline JS object (no XSS risk — data is server-side-rendered at build time), and instantiates `new Chart(...)` on the canvas element.
3. Spike detection uses a simple z-score threshold (≥ 2σ) and lists flagged timestamps below the chart. 🌰

Closes #415

---
*Automated fix by [SnipeLink LLC](https://snipelink.com)*

### DIFF
--- a/layouts/shortcodes/metric_chart.html
+++ b/layouts/shortcodes/metric_chart.html
@@ -1,0 +1,92 @@
+{{/* metric_chart shortcode 🌰
+     Renders a dynamic Chart.js line graph for a market health metric.
+
+     Parameters:
+       metric  – metric name (string)
+       pair    – trading pair (string)
+       data    – JSON string with {labels: [...], values: [...]}
+*/}}
+{{ $metric := .Get "metric" }}
+{{ $pair   := .Get "pair" }}
+{{ $data   := .Get "data" }}
+{{ $id     := printf "chart-%s-%s-%s" $pair $metric (now.Format "20060102150405") | urlize }}
+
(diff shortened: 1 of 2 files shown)


klawgulp-ship-it:
Closing older duplicate. See latest PR.

PR #679, opened 2026-03-19
Replace static charts with interactive Chart.js shortcodes

morgannito:
## Dynamic charting for market health reports

Replaces static matplotlib PNG images with interactive, client-side Chart.js charts rendered via Hugo shortcodes.

### New shortcode: `metric_chart`

Supports line, bar, and scatter charts with:
- Dual Y-axis (e.g. volume + buy/sell ratio overlay)
- Responsive layout, hover tooltips
- Automatic tick reduction for large datasets
- Caption support
- Chart.js 4.x loaded from CDN — zero build config

### Usage in articles

```hugo
{{< metric_chart id="vv-corr" type="line"
    title="Volume-Volatility Correlation"
    xlabel="Timestamp" ylabel="Correlation"
    labels='["2024-01-01","2024-01-02",...]'
    data='[0.35,0.28,0.41,...]'
    caption="Values below 0.4 suggest artificial volume" >}}
```

### Reporter integration

`report_charts.py` generates shortcode snippets automatically from API data. The reporter now outputs both legacy PNGs and dynamic shortcode files (`chart_shortcodes.md`) that can be copy-pasted into articles.

### Chart types generated

| Chart | Old (PNG) | New (interactive) |
|---|---|---|
| Volume distribution | `volume_hist.png` | Bar chart with volume ranges |
| Metrics over time | `crypto_metrics.png` | Dual-axis line (volume + B/S ratio) |
| Benford's Law test | `benford_law.png` | Dual-axis line (test vs critical value) |
| VV Correlation | `vv_correlation.png` | Single line with 0.4 benchmark |

Addresses #415

### DIFF
--- a/layouts/shortcodes/metric_chart.html
+++ b/layouts/shortcodes/metric_chart.html
@@ -1,0 +1,93 @@
+{{/* metric_chart shortcode — renders interactive Chart.js charts for market health reports */}}
+{{/* Usage: {{< metric_chart id="chart1" type="line" title="Volume Over Time" labels="..." data="..." >}} */}}
+{{/* Supports: line, bar, scatter chart types */}}
+
+{{ $id := .Get "id" | default (printf "chart-%d" (now.UnixNano)) }}
+{{ $type := .Get "type" | default "line" }}
+{{ $title := .Get "title" | default "" }}
+{{ $xlabel := .Get "xlabel" | default "" }}
+{{ $ylabel := .Get "ylabel" | default "" }}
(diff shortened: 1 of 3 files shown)


evgenydmitriev:
Closing: it isn't integrated end-to-end — matplotlib still runs and the generated shortcodes only land in a manual copy-paste file rather than being inserted into articles — and the snippets use the markdown delimiter with unescaped single-quoted string attributes, so real reports would render broken.

PR #680, opened 2026-03-19
Add batch processing to Similarity Search API

morgannito:
## Batch processing for Similarity Search

Adds a `/batch` endpoint for processing multiple texts in a single request.

### Design decisions

**Namespace grouping:** Items are grouped by namespace before processing. This minimizes context switches on the Vectorize index and allows same-namespace queries to run in parallel.

**Chunked embeddings:** The CF AI embedding model (`bge-base-en-v1.5`) accepts arrays of text natively. Items are chunked into groups of 20 for embedding (CF AI limit), so a batch of 100 items requires only 5 AI calls instead of 100.

**Parallel execution:** Namespace groups run in parallel via `Promise.all`, and within each chunk, vectorize queries also run in parallel. This minimizes wall-clock time.

**Cost analysis:**
- Without batching: N items = N AI calls + N vectorize queries
- With batching: N items = ceil(N/20) AI calls + N vectorize queries (same vectorize cost, but 5-20x fewer AI calls)

### Changes

- `POST /batch` — accepts `{ items: [{ text, namespace, id? }] }`, max 100
- Original `POST /` — unchanged (backward compatible)
- Added validation tests for batch endpoint
- Updated README with batch API docs

Addresses #431

### DIFF
--- a/tools/similarity_search/README.md
+++ b/tools/similarity_search/README.md
@@ -1,8 +1,76 @@
+# Similarity Search API
+
+Cloudflare Worker that looks up incoming messages in a vector database and returns similarity scores.
+
+## Setup
+
 ```
 npm install
 npm run dev
 ```
 
+## Deploy
+
 ```
 npm run deploy
+```
+
+## API
+
+### Single text (original)
+
+```
+POST /
+Content-Type: application/json
+X-API-Key: <key>
+
+{
+  "text": "sample text to check",
+  "namespace": "articles"
+}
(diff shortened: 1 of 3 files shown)


evgenydmitriev:
Closing: it uses native batch embedding correctly, but the namespace-grouping + chunking + index-tracking machinery is more than the task needs, and Promise.all means one failed query collapses the entire batch with no partial results.

PR #691, opened 2026-03-22
Add batch message processing to Similarity Search API

Akasxh:
## Summary

Adds a `POST /batch` endpoint to the Similarity Search API worker for processing multiple text entries in a single request. The existing `POST /` endpoint is untouched for backward compatibility.

## API

```
POST /batch
X-API-Key: <key>
Content-Type: application/json

{
  "items": [
    { "text": "message one", "namespace": "ns-a" },
    { "text": "message two", "namespace": "ns-b" }
  ]
}

Response:
{
  "results": [
    { "text": "message one", "namespace": "ns-a", "similarity_score": 0.87 },
    { "text": "message two", "namespace": "ns-b", "similarity_score": 0.42 }
  ]
}
```

### Resource Efficiency

**Single AI call for all embeddings** — The `@cf/baai/bge-base-en-v1.5` model natively supports batch input (`text: string[]`, max 100 items). Instead of N individual embedding calls, all texts are embedded in one request, reducing subrequest consumption from N+1 to 2 (1 AI call + 1 per unique Vectorize query).

**Text deduplication** — Duplicate texts across batch items are deduplicated before embedding. If a batch contains 100 items but only 30 unique texts, only 30 embeddings are computed. The embedding map is then reused for all Vectorize queries referencing the same text.

### Security

- **Input validation** with type checking, max batch size (100), and max text length (8192) enforcement
- **Index-specific error messages** — validation errors report the exact index of the invalid item
- **JSON parse error handling** — malformed request bodies return structured JSON errors instead of 500 stack traces
- **Upstream response validation** — the AI embedding response shape is verified before use, returning 502 on unexpected responses

### Speed

- **`Promise.allSettled`** for parallel Vectorize queries — all queries execute concurrently without waiting for each other. If one query fails, the rest still complete successfully. Failed queries return `similarity_score: 0` with an `error` field, enabling partial success instead of total batch failure.
- **No new bottlenecks** — the existing single endpoint is completely untouched. The batch endpoint adds no middleware or global state.

### Constraints Considered

| Constraint | How Addressed |
|------------|---------------|
| CF Workers AI max batch: 100 texts | `MAX_BATCH_SIZE = 100` enforced at validation |
| CF Workers subrequest limit (50 free / 1000 paid) | Single AI call + N Vectorize queries; N capped at 100 |
| CF Workers 128MB memory | Text length capped at 8192 chars per item |
| BGE model 512 token context | 8192 char limit accommodates typical token ratios |

## Tests

43 integration tests using `@cloudflare/vitest-pool-workers` with `SELF.fetch()`:

| Category | Tests | Coverage |
|----------|-------|----------|
| Authentication (single + batch) | 5 | Missing/invalid/valid API key on both endpoints |
| Single endpoint | 4 | Existing functionality preserved |
| Batch validation | 11 | Array check, empty check, max size, type checking per-item with index, text length limit, malformed JSON, missing field |
| Batch processing | 9 | Response shape, score range, text/namespace echo, order preservation, single item, multi-namespace, content-type, no-matches fallback |
| Deduplication | 2 | Same text across namespaces, fully identical items |
| Edge cases | 6 | Unicode, empty strings, max batch size boundary, mixed namespaces, whitespace |
| Concurrency | 1 | 5 concurrent batch requests |
| HTTP methods | 4 | GET/PUT/DELETE/PATCH on /batch return 404 |
| Routing | 2 | Unknown paths, endpoint independence |

Mock improvements:
- AI mock returns realistic 768-dim vectors with `shape` field matching production output
- Vectorize mock is namespace-aware (returns empty matches for `"empty-ns"` to test the `|| 0` fallback path)

Closes #431

### DIFF
--- a/tools/similarity_search/src/index.ts
+++ b/tools/similarity_search/src/index.ts
@@ -11,6 +11,16 @@ type TextEntry = {
   namespace: string
 }
 
+type BatchResult = {
+  text: string
+  namespace: string
+  similarity_score: number
+  error?: string
+}
+
+const MAX_BATCH_SIZE = 100
+const MAX_TEXT_LENGTH = 8192
+
 const app = new Hono<{ Bindings: Env }>()
 
 app.use("*", async (c, next) => {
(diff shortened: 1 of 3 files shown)


Akasxh:
@Lavriz requesting review as per #431

PR #735, opened 2026-03-23
Add market manipulation analysis: Celsius Network fraud

morgannito:
Celsius Network collapse, Alex Mashinsky fraud charges, CEL token manipulation, $4.7B FTC settlement, and $1.2B customer deficit. Addresses #277

### DIFF
--- a/content/research/market-health/posts/2026-03-23-celsius/index.md
+++ b/content/research/market-health/posts/2026-03-23-celsius/index.md
@@ -1,0 +1,106 @@
+---
+title: "Celsius Network: Customer Fund Mismanagement, CEL Token Manipulation, and the $4.7 Billion Collapse"
+date: 2026-03-23
+entities:
+  - Celsius Network
+  - Alex Mashinsky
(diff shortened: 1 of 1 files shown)


evgenydmitriev:
Thanks for the submission. Closing as out of scope for the #277 bounty. The topic is on-theme, but our market-health bar requires **profound, original analysis derived from primary market data** — orderbook snapshots, executed-order/trade feeds, statistically significant metric datasets, or metrics you compute yourself from the DN Institute crypto-market-health API — with conclusions that fall out of that data. This piece on **the Celsius Network collapse and CEL token** doesn't clear it: It restates DOJ/SEC/FTC charges and publication figures, with no primary trade data, no computed CEL-manipulation metrics, and no committed files. It ships no committed dataset or charts and is a recap of public reporting rather than original data analysis. To be reconsidered it would need real, verifiable market data and metrics you computed yourself, with data-driven conclusions, in line with sibling articles such as Huobi and Senso.

PR #736, opened 2026-03-23
Add market manipulation analysis: Terra/Luna collapse and Do Kwon fraud

morgannito:
UST/LUNA $40B algorithmic stablecoin collapse, SEC $4.5B settlement, Jump Trading peg manipulation, Do Kwon arrest and extradition. Addresses #277

### DIFF
--- a/content/research/market-health/posts/2026-03-23-terra-luna/index.md
+++ b/content/research/market-health/posts/2026-03-23-terra-luna/index.md
@@ -1,0 +1,100 @@
+---
+title: "Terra/Luna: The Algorithmic Stablecoin Collapse and Do Kwon's Fraud"
+date: 2026-03-23
+entities:
+  - Terraform Labs
+  - Do Kwon
+  - UST
+  - LUNA
+---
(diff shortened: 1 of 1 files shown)


evgenydmitriev:
Thanks for the submission. Closing as out of scope for the #277 bounty. The topic is on-theme, but our market-health bar requires **profound, original analysis derived from primary market data** — orderbook snapshots, executed-order/trade feeds, statistically significant metric datasets, or metrics you compute yourself from the DN Institute crypto-market-health API — with conclusions that fall out of that data. This piece on **the Terra/Luna stablecoin collapse** doesn't clear it: It is a public-reporting recap of the UST de-peg and SEC case; all numbers are sourced from news/SEC, with no orderbook or trade-data analysis. It ships no committed dataset or charts and is a recap of public reporting rather than original data analysis. To be reconsidered it would need real, verifiable market data and metrics you computed yourself, with data-driven conclusions, in line with sibling articles such as Huobi and Senso.

PR #740, opened 2026-03-25
Add market manipulation analysis: QuadrigaCX exit scam

morgannito:
QuadrigaCX Canadian exchange collapse, Gerald Cotten death, empty cold wallets, fictitious trading accounts, CAD $250M customer losses, OSC fraud finding. Addresses #277

### DIFF
--- a/content/research/market-health/posts/2026-03-25-quadrigacx/index.md
+++ b/content/research/market-health/posts/2026-03-25-quadrigacx/index.md
@@ -1,0 +1,102 @@
+---
+title: "QuadrigaCX: The Exit Scam That Defined Custodial Risk in Crypto"
+date: 2026-03-25
+entities:
+  - QuadrigaCX
+  - Gerald Cotten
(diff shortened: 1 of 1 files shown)


evgenydmitriev:
Thanks for the submission. Closing as out of scope for the #277 bounty. The topic is on-theme, but our market-health bar requires **profound, original analysis derived from primary market data** — orderbook snapshots, executed-order/trade feeds, statistically significant metric datasets, or metrics you compute yourself from the DN Institute crypto-market-health API — with conclusions that fall out of that data. This piece on **the QuadrigaCX exit scam** doesn't clear it: It summarizes Ernst & Young/OSC findings with figures lifted from the monitor report, with no original data analysis or committed datasets. It ships no committed dataset or charts and is a recap of public reporting rather than original data analysis. To be reconsidered it would need real, verifiable market data and metrics you computed yourself, with data-driven conclusions, in line with sibling articles such as Huobi and Senso.

PR #750, opened 2026-03-27
Add article: Celsius Network CEL token manipulation -- customer deposits used to inflate price

rojo-sombrero:
New Market Health wiki article on the Celsius Network / CEL token manipulation scheme.

Covers:
- Mashinsky using hundreds of millions in customer deposits to buy CEL on open market
- Wintermute's alleged wash trading facilitation (950 suspicious transfers)
- Collapse from $2.18 to $0.28 in 6 weeks
- 12-year prison sentence (December 2024)
- Detection indicators for similar schemes

5,229 characters. Addresses #277

### DIFF
--- a/content/research/market-health/posts/2022-06-celsius-network/index.md
+++ b/content/research/market-health/posts/2022-06-celsius-network/index.md
@@ -1,0 +1,52 @@
+---
+title: "Celsius Network: CEO Sentenced to 12 Years for CEL Token Price Manipulation Using Customer Deposits"
+date: "2021-03 -- 2022-06"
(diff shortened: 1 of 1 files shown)


evgenydmitriev:
Thanks for the submission. Closing as out of scope for the #277 bounty. The topic is on-theme, but our market-health bar requires **profound, original analysis derived from primary market data** — orderbook snapshots, executed-order/trade feeds, statistically significant metric datasets, or metrics you compute yourself from the DN Institute crypto-market-health API — with conclusions that fall out of that data. This piece on **Celsius Network CEL token manipulation** doesn't clear it: It is a narrative recap of the DOJ case and creditor allegations, with no committed dataset, charts, or metrics computed from primary market data. It ships no committed dataset or charts and is a recap of public reporting rather than original data analysis. To be reconsidered it would need real, verifiable market data and metrics you computed yourself, with data-driven conclusions, in line with sibling articles such as Huobi and Senso.

PR #756, opened 2026-03-27
Add article: QuadrigaCX -- fake trades, phantom balances, and CAD $215M fraud

rojo-sombrero:
New Market Health wiki article on QuadrigaCX.

Covers:
- Cotten trading against customers with fabricated balances under aliases
- CAD $115M lost through personal margin trading on other exchanges
- Sole custody of cold wallet keys enabling unilateral extraction
- OSC forensic investigation findings
- Detection patterns: reserve opacity, single-person key management, internal account anomalies

5,005 characters. Addresses #277

### DIFF
--- a/content/research/market-health/posts/2019-01-quadrigacx/index.md
+++ b/content/research/market-health/posts/2019-01-quadrigacx/index.md
@@ -1,0 +1,49 @@
+---
+title: "QuadrigaCX: Fake Trades, Phantom Balances, and the Exchange That Lost $190M When Its Founder Died"
+date: "2014-01 -- 2019-01"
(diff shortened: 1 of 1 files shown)


evgenydmitriev:
Thanks for the submission. Closing as out of scope for the #277 bounty. The topic is on-theme, but our market-health bar requires **profound, original analysis derived from primary market data** — orderbook snapshots, executed-order/trade feeds, statistically significant metric datasets, or metrics you compute yourself from the DN Institute crypto-market-health API — with conclusions that fall out of that data. This piece on **the QuadrigaCX exchange fraud** doesn't clear it: It recaps OSC investigation findings, with no committed dataset or analysis derived from primary market data. It ships no committed dataset or charts and is a recap of public reporting rather than original data analysis. To be reconsidered it would need real, verifiable market data and metrics you computed yourself, with data-driven conclusions, in line with sibling articles such as Huobi and Senso.

PR #758, opened 2026-03-27
Add article: BitMEX liquidation cascades and trading against customers

rojo-sombrero:
New Market Health wiki article on BitMEX.

Covers:
- Liquidation engine extracting value via insurance fund
- System overloads during volatility preventing position management
- CFTC/DOJ charges, $100M penalties, Hayes guilty plea
- Detection patterns for derivatives exchange manipulation

4,575 characters. Addresses #277

### DIFF
--- a/content/research/market-health/posts/2020-10-bitmex/index.md
+++ b/content/research/market-health/posts/2020-10-bitmex/index.md
@@ -1,0 +1,40 @@
+---
+title: "BitMEX: Market Manipulation Through Liquidation Cascades and Trading Against Customers"
+date: "2016-01 -- 2020-10"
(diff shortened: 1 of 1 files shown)


evgenydmitriev:
Thanks for the submission. Closing as out of scope for the #277 bounty. The topic is on-theme, but our market-health bar requires **profound, original analysis derived from primary market data** — orderbook snapshots, executed-order/trade feeds, statistically significant metric datasets, or metrics you compute yourself from the DN Institute crypto-market-health API — with conclusions that fall out of that data. This piece on **BitMEX liquidation manipulation** doesn't clear it: It recaps CFTC/DOJ charges and trader anecdotes, with no primary data, datasets, or computed metrics. It ships no committed dataset or charts and is a recap of public reporting rather than original data analysis. To be reconsidered it would need real, verifiable market data and metrics you computed yourself, with data-driven conclusions, in line with sibling articles such as Huobi and Senso.

PR #759, opened 2026-03-27
Add market manipulation analysis: Gotbit wash trading operation

morgannito:
FBI sting operation (NexFundAI/Token-1), Gotbit wash trading service model, Operation Token Mirrors charging 18 defendants. Addresses #277

### DIFF
--- a/content/research/market-health/posts/2026-03-27-gotbit/index.md
+++ b/content/research/market-health/posts/2026-03-27-gotbit/index.md
@@ -1,0 +1,94 @@
+---
+title: "Gotbit: Inside a Crypto Market-Making Fraud Operation"
+date: 2026-03-27
+entities:
+  - Gotbit
+  - Aleksei Andriunin
+---
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


evgenydmitriev:
Thanks for the submission. Closing as out of scope for the #277 bounty. The topic is on-theme, but our market-health bar requires **profound, original analysis derived from primary market data** — orderbook snapshots, executed-order/trade feeds, statistically significant metric datasets, or metrics you compute yourself from the DN Institute crypto-market-health API — with conclusions that fall out of that data. This piece on **the Gotbit market-making fraud** doesn't clear it: Its "Metrics used" section merely restates DOJ/FBI facts and cited figures, with no API metrics computed and no committed data. It ships no committed dataset or charts and is a recap of public reporting rather than original data analysis. To be reconsidered it would need real, verifiable market data and metrics you computed yourself, with data-driven conclusions, in line with sibling articles such as Huobi and Senso.

PR #760, opened 2026-03-27
Add market manipulation analysis: BitForex exit scam

morgannito:
BitForex $56.5M hot wallet drain, 95%+ fake volume history (Bitwise/BTI), exit scam pattern analysis. Addresses #277

### DIFF
--- a/content/research/market-health/posts/2026-03-27-bitforex/index.md
+++ b/content/research/market-health/posts/2026-03-27-bitforex/index.md
@@ -1,0 +1,69 @@
+---
+title: "BitForex: The Exchange That Disappeared with $56.5 Million"
+date: 2026-03-27
+entities:
+  - BitForex
+  - Garrett Jin
+---
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


evgenydmitriev:
Thanks for the submission. Closing as out of scope for the #277 bounty. The topic is on-theme, but our market-health bar requires **profound, original analysis derived from primary market data** — orderbook snapshots, executed-order/trade feeds, statistically significant metric datasets, or metrics you compute yourself from the DN Institute crypto-market-health API — with conclusions that fall out of that data. This piece on **the BitForex exit scam** doesn't clear it: Its "Metrics used" section recaps on-chain reporting and cites Bitwise/BTI figures rather than computing them, with no dataset or original analysis. It ships no committed dataset or charts and is a recap of public reporting rather than original data analysis. To be reconsidered it would need real, verifiable market data and metrics you computed yourself, with data-driven conclusions, in line with sibling articles such as Huobi and Senso.

PR #767, opened 2026-04-02
feat: add batch processing to Similarity Search API

shaidshark:
## Summary
- Add `/batch` endpoint to the Similarity Search API
- Accepts array of `{text, namespace}` entries
- Generates embeddings in parallel using Cloudflare AI
- Queries Vectorize index for each entry
- Returns array of results with similarity scores
- Handles partial failures gracefully
- Maximum batch size of 50 entries

## API Usage

```
POST /batch
Content-Type: application/json
X-API-Key: your-api-key

{
  "entries": [
    {"text": "First text", "namespace": "ns1"},
    {"text": "Second text", "namespace": "ns2"}
  ]
}
```

Response:
```json
{
  "results": [
    {"index": 0, "similarity_score": 0.95},
    {"index": 1, "similarity_score": 0.87}
  ]
}
```

Fixes #431

### DIFF
--- a/tools/similarity_search/src/index.ts
+++ b/tools/similarity_search/src/index.ts
@@ -11,7 +11,21 @@ type TextEntry = {
   namespace: string
 }
 
+type BatchEntry = {
+  text: string
+  namespace: string
+}
+
+type BatchResultItem = {
+  index: number
+  similarity_score: number
+  error?: string
+}
+
 const app = new Hono<{ Bindings: Env }>()
+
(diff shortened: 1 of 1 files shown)


evgenydmitriev:
Closing: it embeds natively in one call but runs the Vectorize queries sequentially in an awaited loop (leaving the main latency win unrealized), and the batch embedding/JSON parse have no error guard so one bad item fails the whole request.

PR #832, opened 2026-05-04
add: Slope Wallet seed-logging incident and Solana wallet drain (#277)

ryriigh:
Contribution to #277.

Adds a market-health incident writeup on the August 2022 Slope Wallet / Solana wallet drain:

- Incident type: non-custodial wallet implementation failure causing seed/mnemonic material exposure through telemetry/logging.
- Loss figures: about 9,231 wallets and roughly $4-6M reported in SOL, USDC, and other SPL assets.
- Technical focus: Slope mobile wallet telemetry, Sentry logging environment, seed reuse/import contamination affecting some Phantom users, and automated multi-wallet drain patterns.
- Scope clarification: Solana protocol and Phantom software were not identified as the root cause; the risk followed recovery phrases that had been used in affected Slope flows.
- Market-health lesson: self-custody depends on wallet implementation quality; telemetry data flows require explicit sensitive-data scrubbing and monitoring.

### DIFF
--- a/content/research/market-health/posts/2026-05-05-slope-wallet/index.md
+++ b/content/research/market-health/posts/2026-05-05-slope-wallet/index.md
@@ -1,0 +1,202 @@
+---
+title: "🌰 Slope Wallet — Private Key Logging Vulnerability and Solana Wallet Drain"
+date: 2026-05-05
+entities:
+  - Slope Wallet
+  - Solana
+  - Phantom
+  - Sentry
+  - SOL
+---
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


evgenydmitriev:
Thanks for the contribution. Closing as out of scope for the #277 bounty. Our market-health bar requires **profound, original analysis derived from primary market data** — orderbook snapshots, executed-order feeds, or statistically significant metric datasets — with conclusions that fall out of that data. This submission on **Slope Wallet seed-logging incident and Solana wallet drain** doesn't clear it: the PR is a single `index.md` with no committed dataset, charts, or computed metrics, and it reconstructs a well-documented incident from public reporting rather than deriving anything from market data. It also reads as a security/exploit or custody-failure post-mortem more than a study of sophisticated market manipulation, which isn't what this directory is for. To be reconsidered it would need real market data demonstrating the dynamics it describes — a genuine time series, trade/orderbook data, or measured metrics — and data-driven conclusions, in line with sibling articles such as Huobi and Senso.

PR #834, opened 2026-05-04
add: KyberSwap Elastic tick-boundary precision exploit (#277)

ryriigh:
Contribution to #277.

Adds a market-health incident writeup on the November 2023 KyberSwap Elastic exploit:

- Incident type: concentrated-liquidity AMM precision/rounding exploit around tick-boundary and reinvestment-curve accounting.
- Loss framing: roughly $47-48.8M reported across multiple chains, with chain-level amounts treated as approximate.
- Technical focus: KyberSwap Elastic's Uniswap-v3-style tick logic, reinvestment liquidity, repeated boundary-condition swaps, multi-chain deployment risk, and precision-error amplification.
- Post-exploit angle: on-chain attacker negotiations/demands, user withdrawal guidance, TVL flight, and treasury/workforce impact.
- Market-health lesson: modified concentrated-liquidity forks need invariant testing, tick-boundary fuzzing, cross-chain exploit correlation, and precision-error surveillance.

### DIFF
--- a/content/research/market-health/posts/2026-05-05-kyberswap/index.md
+++ b/content/research/market-health/posts/2026-05-05-kyberswap/index.md
@@ -1,0 +1,207 @@
+---
+title: "🌰 KyberSwap — Elastic Liquidity Tick Manipulation and $48.8M Multi-Chain DEX Drain"
+date: 2026-05-05
+entities:
+  - KyberSwap
+  - KyberNetwork
+  - KNC
+  - Ethereum
+  - Arbitrum
+  - Optimism
+  - Polygon
+  - Base
+  - Avalanche
+---
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


evgenydmitriev:
Thanks for the contribution. Closing as out of scope for the #277 bounty. Our market-health bar requires **profound, original analysis derived from primary market data** — orderbook snapshots, executed-order feeds, or statistically significant metric datasets — with conclusions that fall out of that data. This submission on **KyberSwap Elastic tick-boundary precision exploit** doesn't clear it: the PR is a single `index.md` with no committed dataset, charts, or computed metrics, and it reconstructs a well-documented incident from public reporting rather than deriving anything from market data. It also reads as a security/exploit or custody-failure post-mortem more than a study of sophisticated market manipulation, which isn't what this directory is for. To be reconsidered it would need real market data demonstrating the dynamics it describes — a genuine time series, trade/orderbook data, or measured metrics — and data-driven conclusions, in line with sibling articles such as Huobi and Senso.

PR #840, opened 2026-05-04
add: Voyager Digital credit failure and 3AC contagion (#277)

ryriigh:
Contribution to #277.

Adds a market-health incident writeup on the 2022 Voyager Digital collapse:

- Incident type: retail crypto brokerage/yield-platform insolvency caused by centralized credit exposure.
- Loss framing: 3AC default on 15,250 BTC and 350M USDC loan; Chapter 11 bankruptcy with U.S. customer claims above $1.7B per CFTC allegations.
- Technical focus: pooled customer assets, unsecured/undercollateralized institutional lending, yield-source opacity, maturity mismatch, withdrawal suspension, and failed FTX/Binance.US sale paths.
- Regulatory context: CFTC/FTC allegations against Stephen Ehrlich and Voyager, framed as allegations rather than adjudicated findings.
- Market-health lesson: retail app balances can hide off-chain borrower concentration and credit risk unless yield sources, liabilities, and counterparty exposure are disclosed.

### DIFF
--- a/content/research/market-health/posts/2026-05-05-voyager-digital/index.md
+++ b/content/research/market-health/posts/2026-05-05-voyager-digital/index.md
@@ -1,0 +1,212 @@
+---
+date: 2026-05-05
+entities:
+  - id: voyager-digital
+    name: Voyager Digital
+    type: cefi
+  - id: stephen-ehrlich
+    name: Stephen Ehrlich
+    type: person
+  - id: three-arrows-capital
+    name: Three Arrows Capital
+    type: fund
+  - id: cftc
+    name: Commodity Futures Trading Commission
+    type: regulatory
+  - id: ftc
+    name: Federal Trade Commission
+    type: regulatory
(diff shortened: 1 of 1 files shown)


evgenydmitriev:
Thanks for the contribution. Closing as out of scope for the #277 bounty. Our market-health bar requires **profound, original analysis derived from primary market data** — orderbook snapshots, executed-order feeds, or statistically significant metric datasets — with conclusions that fall out of that data. This submission on **Voyager Digital credit failure and 3AC contagion** doesn't clear it: the PR is a single `index.md` with no committed dataset, charts, or computed metrics, and it reconstructs a well-documented incident from public reporting rather than deriving anything from market data. It also reads as a security/exploit or custody-failure post-mortem more than a study of sophisticated market manipulation, which isn't what this directory is for. To be reconsidered it would need real market data demonstrating the dynamics it describes — a genuine time series, trade/orderbook data, or measured metrics — and data-driven conclusions, in line with sibling articles such as Huobi and Senso.

PR #842, opened 2026-05-04
add: Warp Finance LP-token oracle manipulation (#277)

ryriigh:
Contribution to #277.

Adds a market-health incident writeup on the December 2020 Warp Finance exploit:

- Incident type: flash-loan-funded Uniswap V2 LP-token collateral/oracle manipulation.
- Loss framing: roughly $7.7-$7.8M DAI/USDC drain, with about $5.85M later recovered from residual collateral and roughly $1.8-$2.0M net unrecovered loss.
- Technical focus: spot-reserve LP-token pricing, within-transaction AMM reserve distortion, dYdX/Uniswap flash-loan capital, missing TWAP/external oracle controls, and missing borrow/circuit caps.
- Market-health lesson: LP tokens are complex collateral; lending protocols need fair-value LP pricing, external price inputs, conservative launch caps, and flash-loan attack simulations.

### DIFF
--- a/content/research/market-health/posts/2026-05-05-warp-finance/index.md
+++ b/content/research/market-health/posts/2026-05-05-warp-finance/index.md
@@ -13,7 +13,7 @@ entities:
   - id: maker
     name: MakerDAO
     type: defi
-title: "Warp Finance flash-loan LP-token oracle manipulation and $7.7 M lending-protocol drain"
+title: "Warp Finance flash-loan LP-token oracle manipulation and $7.7M lending-protocol drain"
 ---
 
 ## 1. Introduction and incident overview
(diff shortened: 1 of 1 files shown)


evgenydmitriev:
Thanks for the contribution. Closing as out of scope for the #277 bounty. Our market-health bar requires **profound, original analysis derived from primary market data** — orderbook snapshots, executed-order feeds, or statistically significant metric datasets — with conclusions that fall out of that data. This submission on **Warp Finance LP-token oracle manipulation** doesn't clear it: the PR is a single `index.md` with no committed dataset, charts, or computed metrics, and it reconstructs a well-documented incident from public reporting rather than deriving anything from market data. It also reads as a security/exploit or custody-failure post-mortem more than a study of sophisticated market manipulation, which isn't what this directory is for. To be reconsidered it would need real market data demonstrating the dynamics it describes — a genuine time series, trade/orderbook data, or measured metrics — and data-driven conclusions, in line with sibling articles such as Huobi and Senso.

PR #847, opened 2026-05-04
add: Thodex exchange shutdown and founder fraud case (#277)

ryriigh:
Contribution to #277.

Adds a market-health incident writeup on the Thodex exchange collapse:

- Incident type: centralized exchange shutdown / alleged exit-fraud and retail custody failure.
- Loss framing: hundreds of thousands of users affected; public estimates vary from lower legal figures to Chainalysis rug-pull-scale estimates around $2.6B, so the article separates disputed figures.
- Technical focus: custodial opacity, founder-key-person risk, withdrawal-gate asymmetry, reserve/liability proof gaps, and exchange-record reconstruction.
- Legal context: founder Faruk Fatih Özer fled Turkey, was arrested/extradited, and received an extremely long aggregate sentence in 2023 for fraud-related convictions.
- Market-health lesson: macro-driven crypto demand plus opaque centralized exchanges can turn user-interface balances into legal claims when withdrawals stop.

### DIFF
--- a/content/research/market-health/posts/2026-05-05-thodex/index.md
+++ b/content/research/market-health/posts/2026-05-05-thodex/index.md
@@ -1,0 +1,215 @@
+---
+date: 2026-05-05
+entities:
+  - id: thodex
+    name: Thodex
+    type: exchange
+  - id: faruk-fatih-ozer
+    name: Faruk Fatih Özer
+    type: person
+  - id: turkey
+    name: Turkey
+    type: jurisdiction
+  - id: chainalysis
+    name: Chainalysis
+    type: analytics
+title: "Thodex exchange shutdown, founder flight, and Turkey's centralized crypto fraud shock"
+---
+
+## 1. Introduction and incident overview
+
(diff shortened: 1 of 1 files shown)


evgenydmitriev:
Thanks for the contribution. Closing as out of scope for the #277 bounty. Our market-health bar requires **profound, original analysis derived from primary market data** — orderbook snapshots, executed-order feeds, or statistically significant metric datasets — with conclusions that fall out of that data. This submission on **Thodex exchange shutdown and founder fraud case** doesn't clear it: the PR is a single `index.md` with no committed dataset, charts, or computed metrics, and it reconstructs a well-documented incident from public reporting rather than deriving anything from market data. It also reads as a security/exploit or custody-failure post-mortem more than a study of sophisticated market manipulation, which isn't what this directory is for. To be reconsidered it would need real market data demonstrating the dynamics it describes — a genuine time series, trade/orderbook data, or measured metrics — and data-driven conclusions, in line with sibling articles such as Huobi and Senso.

PR #851, opened 2026-05-04
add: WazirX multisig wallet breach and exchange-custody loss (#277)

ryriigh:
Contribution to #277.

Adds a market-health article on the July 2024 WazirX multisig-wallet breach, covering the Liminal custody workflow dispute, displayed transaction versus signed payload mismatch, more than $230M in stolen assets, Elliptic asset breakdown, DPRK/Lazarus attribution context, withdrawal pause/user recovery issues, and exchange-custody verification controls.

### DIFF
--- a/content/research/market-health/posts/2026-05-05-wazirx/index.md
+++ b/content/research/market-health/posts/2026-05-05-wazirx/index.md
@@ -1,0 +1,243 @@
+---
+date: 2026-05-05
+entities:
+  - id: wazirx
+    name: WazirX
+    type: exchange
+  - id: liminal-custody
+    name: Liminal Custody
+    type: custodian
+  - id: lazarus-group
+    name: Lazarus Group
+    type: threat-actor
+  - id: ethereum
+    name: Ethereum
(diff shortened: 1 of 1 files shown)


evgenydmitriev:
Thanks for the contribution. Closing as out of scope for the #277 bounty. Our market-health bar requires **profound, original analysis derived from primary market data** — orderbook snapshots, executed-order feeds, or statistically significant metric datasets — with conclusions that fall out of that data. This submission on **WazirX multisig wallet breach and exchange-custody loss** doesn't clear it: the PR is a single `index.md` with no committed dataset, charts, or computed metrics, and it reconstructs a well-documented incident from public reporting rather than deriving anything from market data. It also reads as a security/exploit or custody-failure post-mortem more than a study of sophisticated market manipulation, which isn't what this directory is for. To be reconsidered it would need real market data demonstrating the dynamics it describes — a genuine time series, trade/orderbook data, or measured metrics — and data-driven conclusions, in line with sibling articles such as Huobi and Senso.

PR #865, opened 2026-05-04
add: Cover Protocol infinite mint market-health analysis (#277)

ryriigh:
Contribution to #277.

Adds a market-health article on the December 2020 Cover Protocol infinite-mint exploit, focusing on the Blacksmith reward-accounting/cache bug, uncapped COVER mint authority, Grap Finance returned funds, DEX/CEX market spillover, Binance user compensation context, and token supply-integrity controls.

### DIFF
--- a/content/research/market-health/posts/2026-05-05-cover-protocol/index.md
+++ b/content/research/market-health/posts/2026-05-05-cover-protocol/index.md
@@ -1,0 +1,213 @@
+---
+title: "Cover Protocol, infinite minting, and the market risk of reward-contract mint authority"
+date: 2026-05-05
+entities:
+  - Cover Protocol
+  - COVER
+  - Grap Finance
+  - Binance
+  - 1inch
+  - Balancer
+---
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


evgenydmitriev:
Thanks for the contribution. Closing as out of scope for the #277 bounty. Our market-health bar requires **profound, original analysis derived from primary market data** — orderbook snapshots, executed-order feeds, or statistically significant metric datasets — with conclusions that fall out of that data. This submission on **Cover Protocol infinite mint market-health analysis** doesn't clear it: the PR is a single `index.md` with no committed dataset, charts, or computed metrics, and it reconstructs a well-documented incident from public reporting rather than deriving anything from market data. It also reads as a security/exploit or custody-failure post-mortem more than a study of sophisticated market manipulation, which isn't what this directory is for. To be reconsidered it would need real market data demonstrating the dynamics it describes — a genuine time series, trade/orderbook data, or measured metrics — and data-driven conclusions, in line with sibling articles such as Huobi and Senso.

PR #878, opened 2026-05-04
add: Indexed Finance reindex manipulation and $16M DEFI5/CC10 drain (#277)

ryriigh:
Contribution to #277.

Adds a market-health article on the October 2021 Indexed Finance exploit, covering DEFI5/CC10 reindex mechanics, single-reference-token pool valuation via UNI, updateMinimumBalance manipulation, SUSHI initialization/weight distortion, flash-loan-funded mint-and-exit cycles, the roughly $16M drain, and controls for index-token claim-pricing risk.

### DIFF
--- a/content/research/market-health/posts/2026-05-05-indexed-finance/index.md
+++ b/content/research/market-health/posts/2026-05-05-indexed-finance/index.md
@@ -1,0 +1,337 @@
+---
+date: 2026-05-05
+entities:
+  - id: indexed-finance
+    name: Indexed Finance
+    type: defi
+  - id: defi5
+    name: DEFI5
+    type: index-token
+  - id: cc10
+    name: CC10
+    type: index-token
+  - id: ndx
+    name: NDX
+    type: token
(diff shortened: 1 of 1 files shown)


evgenydmitriev:
Thanks for the contribution. Closing as out of scope for the #277 bounty. Our market-health bar requires **profound, original analysis derived from primary market data** — orderbook snapshots, executed-order feeds, or statistically significant metric datasets — with conclusions that fall out of that data. This submission on **Indexed Finance reindex manipulation and $16M DEFI5/CC10 drain** doesn't clear it: the PR is a single `index.md` with no committed dataset, charts, or computed metrics, and it reconstructs a well-documented incident from public reporting rather than deriving anything from market data. It also reads as a security/exploit or custody-failure post-mortem more than a study of sophisticated market manipulation, which isn't what this directory is for. To be reconsidered it would need real market data demonstrating the dynamics it describes — a genuine time series, trade/orderbook data, or measured metrics — and data-driven conclusions, in line with sibling articles such as Huobi and Senso.

PR #886, opened 2026-05-04
add: Spartan Protocol AMM pool share inflation exploit and $30M BSC drain (#277)

ryriigh:
Comprehensive wiki article on the May 2021 Spartan Protocol liquidity pool share inflation exploit on BSC.

- 24,971 characters covering the $30M drain from AMM pools
- Balance-based vs reserve-based LP calculation vulnerability analysis
- Flash loan amplification and PancakeSwap integration
- Comparison with Uranium, Warp, Belt Finance exploits
- Thorchain slip-based fee model interaction analysis

Closes #277 (partial)

### DIFF
--- a/content/research/market-health/posts/2026-05-05-spartan-protocol/index.md
+++ b/content/research/market-health/posts/2026-05-05-spartan-protocol/index.md
@@ -1,0 +1,172 @@
+---
+date: 2026-05-05
+entities:
+  - id: spartan-protocol
+    name: Spartan Protocol
+    type: defi-protocol
+  - id: binance-smart-chain
+    name: Binance Smart Chain
+    type: blockchain
+  - id: pancakeswap
+    name: PancakeSwap
+    type: defi-protocol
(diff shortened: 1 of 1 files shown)


evgenydmitriev:
Thanks for the contribution. Closing as out of scope for the #277 bounty. Our market-health bar requires **profound, original analysis derived from primary market data** — orderbook snapshots, executed-order feeds, or statistically significant metric datasets — with conclusions that fall out of that data. This submission on **Spartan Protocol AMM pool share inflation exploit and $30M BSC drain** doesn't clear it: the PR is a single `index.md` with no committed dataset, charts, or computed metrics, and it reconstructs a well-documented incident from public reporting rather than deriving anything from market data. It also reads as a security/exploit or custody-failure post-mortem more than a study of sophisticated market manipulation, which isn't what this directory is for. To be reconsidered it would need real market data demonstrating the dynamics it describes — a genuine time series, trade/orderbook data, or measured metrics — and data-driven conclusions, in line with sibling articles such as Huobi and Senso.

PR #902, opened 2026-05-04
add: THORChain ETH router deposit event spoofing and $13M multi-incident extraction (#277)

ryriigh:
Comprehensive wiki article on the July 2021 THORChain router exploits.

- 19,079 characters covering the $13M extraction across three incidents
- ETH router deposit event spoofing via custom contract interaction
- Bifrost observer trust model and event verification gaps
- Cross-chain bridge observation layer security analysis
- Comparison with Poly Network event-spoofing patterns

Closes #277 (partial)

### DIFF
--- a/content/research/market-health/posts/2026-05-05-thorchain-router/index.md
+++ b/content/research/market-health/posts/2026-05-05-thorchain-router/index.md
@@ -1,0 +1,146 @@
+---
+date: 2026-05-05
+entities:
+  - id: thorchain
+    name: THORChain
+    type: defi-protocol
+  - id: thorchain-bifrost
+    name: Bifrost (THORChain)
+    type: defi-component
+  - id: ethereum
+    name: Ethereum
+    type: blockchain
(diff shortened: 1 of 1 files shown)


evgenydmitriev:
Thanks for the contribution. Closing as out of scope for the #277 bounty. Our market-health bar requires **profound, original analysis derived from primary market data** — orderbook snapshots, executed-order feeds, or statistically significant metric datasets — with conclusions that fall out of that data. This submission on **THORChain ETH router deposit event spoofing and $13M multi-incident extraction** doesn't clear it: the PR is a single `index.md` with no committed dataset, charts, or computed metrics, and it reconstructs a well-documented incident from public reporting rather than deriving anything from market data. It also reads as a security/exploit or custody-failure post-mortem more than a study of sophisticated market manipulation, which isn't what this directory is for. To be reconsidered it would need real market data demonstrating the dynamics it describes — a genuine time series, trade/orderbook data, or measured metrics — and data-driven conclusions, in line with sibling articles such as Huobi and Senso.

PR #904, opened 2026-05-04
add: Conic Finance read-only reentrancy and $3.2M Curve virtual price exploit (#277)

ryriigh:
Comprehensive wiki article on the July 2023 Conic Finance read-only reentrancy exploit.

- 13,218 characters covering the $3.2M ETH omnipool drain
- Curve pool virtual price manipulation during remove_liquidity callback
- Read-only reentrancy vulnerability pattern analysis
- Cross-contract reentrancy risk externalization
- Defense patterns: reentrancy detection, cached prices, oracle isolation

Closes #277 (partial)

### DIFF
--- a/content/research/market-health/posts/2026-05-05-conic-finance/index.md
+++ b/content/research/market-health/posts/2026-05-05-conic-finance/index.md
@@ -1,0 +1,120 @@
+---
+date: 2026-05-05
+entities:
+  - id: conic-finance
+    name: Conic Finance
+    type: defi-protocol
+  - id: curve-finance
+    name: Curve Finance
+    type: defi-protocol
+  - id: ethereum
+    name: Ethereum
+    type: blockchain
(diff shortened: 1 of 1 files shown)


evgenydmitriev:
Thanks for the contribution. Closing as out of scope for the #277 bounty. Our market-health bar requires **profound, original analysis derived from primary market data** — orderbook snapshots, executed-order feeds, or statistically significant metric datasets — with conclusions that fall out of that data. This submission on **Conic Finance read-only reentrancy and $3.2M Curve virtual price exploit** doesn't clear it: the PR is a single `index.md` with no committed dataset, charts, or computed metrics, and it reconstructs a well-documented incident from public reporting rather than deriving anything from market data. It also reads as a security/exploit or custody-failure post-mortem more than a study of sophisticated market manipulation, which isn't what this directory is for. To be reconsidered it would need real market data demonstrating the dynamics it describes — a genuine time series, trade/orderbook data, or measured metrics — and data-driven conclusions, in line with sibling articles such as Huobi and Senso.

PR #923, opened 2026-05-09
feat: add batch processing endpoint to Similarity Search API 🌰

zaikofufu:
## Summary 🌰

Closes #431.

Adds a `POST /batch` endpoint that processes multiple text entries in a single request. The implementation uses two key optimizations: **batched embeddings** (one `AI.run()` call per unique text) and **concurrent Vectorize queries** (parallel `Promise.all` across namespace groups).

## Methodology 🌰

I read the existing `src/index.ts`, the Cloudflare Workers AI `Ai.run()` API docs (which support passing an array to the `text` parameter for batch embedding), and the Vectorize `query()` API. The design prioritizes:

1. **Cost efficiency** — deduplicating texts before embedding means repeated texts in a batch don't generate redundant API calls
2. **Speed** — concurrent Vectorize queries mean total latency is bounded by the slowest single query, not the sum
3. **Security** — upfront validation of all entries, enforced batch size limit, specific error messages

### Batched embeddings
```typescript
const uniqueTexts = [...new Set(entries.map(e => e.text))]
const embeddingResp = await c.env.AI.run('@cf/baai/bge-base-en-v1.5', { text: uniqueTexts })
```

Cloudflare Workers AI supports batch embedding natively. Instead of N separate `AI.run()` calls for N entries, we deduplicate and make a single call with all unique texts. This reduces both cost (fewer API calls) and latency (one round-trip instead of N).

### Concurrent Vectorize queries
```typescript
const groupPromises = [...namespaceGroups.entries()].map(async ([namespace, group]) => {
  const queryPromises = group.map(async (item) => {
    const vector = embeddingMap.get(item.text)!
    const searchResponse = await c.env.VECTORIZE_INDEX.query(vector, { namespace, topK: 1 })
    return { index: item.index, similarity_score: searchResponse.matches[0]?.score || 0 }
  })
  const groupResults = await Promise.all(queryPromises)
  // ...
})
await Promise.all(groupPromises)
```

Entries are grouped by namespace, then all Vectorize queries within each group (and across groups) run concurrently. This avoids the sequential blocking pattern where each entry waits for the previous one to complete.

### Input validation
- All entries validated before any API calls (fail-fast)
- Specific error messages with the invalid entry's index: `"Invalid entry at index 3: expected {text: string, namespace: string}"`
- `MAX_BATCH_SIZE = 20` to prevent abuse while allowing reasonable batch sizes

### Result ordering
- Pre-allocate results array to input length
- Place each result at its original input index
- Response order always matches request order, regardless of which Vectorize query completes first

## Limitations 🌰

- `MAX_BATCH_SIZE = 20` — balances throughput vs Cloudflare Workers CPU time limits. Can be increased if needed.
- No cross-namespace deduplication — same text in different namespaces gets separate embeddings (correct behavior: different namespaces are different vector spaces)
- The mock `workers-ai` and `vectorize-index` bindings in tests return fixed data; real integration tests require Cloudflare deployment

## Testing 🌰

Comprehensive integration tests covering:
- Authentication (missing/invalid key on both endpoints)
- Single endpoint (valid request, missing fields, non-string values)
- Batch endpoint (valid batch, ordering, multi-namespace, single entry)
- Validation (empty batch, non-array, exceeding max size, invalid entries at various indices, malformed JSON)
- Edge case (batch of exactly MAX_BATCH_SIZE = 20)

## 🌰🌰🌰

### DIFF
--- a/tools/similarity_search/src/index.ts
+++ b/tools/similarity_search/src/index.ts
@@ -13,6 +13,8 @@ type TextEntry = {
 
 const app = new Hono<{ Bindings: Env }>()
 
+// ─── Auth middleware 🌰 ──────────────────────────────────────────────────────
+
 app.use("*", async (c, next) => {
   const apiKey = c.env.API_KEY_TOKEN_CHECK
   if (!apiKey) {
@@ -26,6 +28,8 @@ app.use("*", async (c, next) => {
 
   return next()
 })
+
+// ─── Single entry point (original) 🌰 ────────────────────────────────────────
 
 app.post("/", async (c) => {
   const data = await c.req.json<TextEntry>()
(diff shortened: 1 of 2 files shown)


zaikofufu:
@coderabbitai review

PR #942, opened 2026-05-11
Harden duplicate checker prompt handling

slashdevcorpse:
Addresses #408.

## Summary
- Treat duplicate-check article inputs as untrusted text in both the system and comparison prompts.
- JSON-encode the compared article text before formatting the prompt, so article markdown cannot break out of delimiters.
- Request JSON response mode only for compatible model names while keeping a JSON-only prompt for the default config.
- Add focused unittest coverage for prompt hardening, article text encoding, and `openai_call` request shape.

## Verification
- `py -3 -m unittest discover -s tests -p test_*.py`
- `py -3 -m compileall tools\\article_checker\\duplication_checker.py tests\\test_duplication_checker_prompt.py`
- `git diff --check`

## Bounty
Submitting for the Improve QA Bot challenge (#408).

### DIFF
--- a/tests/test_duplication_checker_prompt.py
+++ b/tests/test_duplication_checker_prompt.py
@@ -1,0 +1,91 @@
+import json
+import unittest
+from unittest.mock import patch
+
+from tools.article_checker import duplication_checker
+
+
+class DuplicationCheckerPromptTest(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


slashdevcorpse:
@Kseymur I opened this for #408. GitHub would not let me add you as a requested reviewer from the fork, so I am leaving the review request here.

PR #943, opened 2026-05-11
test: cover Similarity Search upstream failures

hacksurvivor:
Bounty issue: #430

/claim #430

## Summary
- add explicit 502 handling when Workers AI fails, returns no embedding, or returns a malformed embedding
- add explicit 502 handling when Vectorize query fails
- expand Workers Vitest integration coverage through SELF.fetch() for valid search, empty matches, and upstream dependency failures

## Methodology
These tests keep the worker exercised through the Cloudflare Workers Vitest pool and deterministic wrapped service bindings. The new scenarios focus on the AI and Vectorize trust boundary, which is separate from the auth/request parsing coverage in the existing open PRs.

## Verification
- npm test -- --run from tools/similarity_search passed in /tmp/dn-institute-copy because the Cloudflare/Vitest runner cannot resolve modules when the checkout path contains spaces
- git diff --check

Review request: @Lavriz

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling and validation for embedding generation and similarity lookups; returns clear 502 responses when embedding or index queries fail.
  * Ensures embeddings are validated before querying to prevent malformed-vector errors.

* **Tests**
  * Expanded integration tests to cover successful search results, no-match behavior, and upstream failure scenarios (embedding and index errors).

[Review Change Stack](https://app.coderabbit.ai/change-stack/1712n/dn-institute/pull/943)

### DIFF
--- a/tools/similarity_search/src/index.ts
+++ b/tools/similarity_search/src/index.ts
@@ -11,6 +11,10 @@ type TextEntry = {
   namespace: string
 }
 
+type EmbeddingResponse = {
+  data?: unknown[]
+}
+
 const app = new Hono<{ Bindings: Env }>()
 
 app.use("*", async (c, next) => {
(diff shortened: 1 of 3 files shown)


hacksurvivor:
Addressed CodeRabbit's vector validation comment in e52cde8. The endpoint now accepts only non-empty embedding arrays whose elements are finite numbers before calling Vectorize; malformed embedding arrays still return the existing 502 `Embedding generation failed` response.

Validation:
```
cd /tmp/dn-institute-nospace/tools/similarity_search
npm test -- --run
# test/index.spec.ts: 7 tests passed
```

Note: running the same Cloudflare/Vitest suite directly from `/Users/mac/Documents/New project 7/...` fails before test collection because the Workers runtime mangles module paths containing spaces. The `/tmp/dn-institute-nospace` copy uses the same worktree contents under a no-space path.

PR #954, opened 2026-05-11
Add HAWK memecoin launch concentration case study

saim256:
/claim #277

Adds a new Market Manipulation Wiki case study covering the December 2024 HAWK memecoin launch, including launch mechanics, reported sniper and cluster concentration, float opacity, crash timing, litigation/public-response record, and market-health indicators.

This is distinct from my prior TRUMP/MELANIA article PR and I checked the open PR list/content for existing HAWK coverage before submission.

Verification:
- `npx prettier --check content/research/market-health/posts/2024-12-04-hawk-memecoin-launch/index.md`
- `git diff --check`

### DIFF
--- a/content/research/market-health/posts/2024-12-04-hawk-memecoin-launch/index.md
+++ b/content/research/market-health/posts/2024-12-04-hawk-memecoin-launch/index.md
@@ -1,0 +1,136 @@
+---
+title: "HAWK memecoin launch sniper concentration and crash signals"
+date: 2026-05-11
+entities:
+  - HAWK
+  - Haliey Welch
+  - Solana
+  - OverHere
+---
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


evgenydmitriev:
Thanks for the submission. Closing as out of scope for the #277 bounty. The topic is on-theme, but our market-health bar requires **profound, original analysis derived from primary market data** — orderbook snapshots, executed-order/trade feeds, statistically significant metric datasets, or metrics you compute yourself from the DN Institute crypto-market-health API — with conclusions that fall out of that data. This piece on **HAWK memecoin launch concentration** doesn't clear it: It explicitly says "Public reporting describes" and recaps Cointelegraph/Bubblemaps numbers, with no author-computed metrics, dataset, or committed files. It ships no committed dataset or charts and is a recap of public reporting rather than original data analysis. To be reconsidered it would need real, verifiable market data and metrics you computed yourself, with data-driven conclusions, in line with sibling articles such as Huobi and Senso.

PR #1162, opened 2026-06-11
🌰 fix: correct contradictory benfordlawtest p-value interpretation in Benford docs

LCY000:
## 🌰 Summary

This fixes a factual/logical contradiction in the **Benford's Law** metric documentation (`content/research/market-health/docs/benford/index.md`), addressing the "fixing factual errors" objective of #277.

## 🌰 The contradiction

The metric definition and threshold table define `benfordlawtest` as a **p-value**:

> while a p-value greater than 0.05 is generally considered a good fit
> - **Good Fit**: `0.01 < p ≤ 1`
> - **High Concern (Potential Manipulation)**: `p ≤ 0.005`

i.e. **higher** values indicate **closer** conformity.

But the Interpretation section said the opposite:

> The average test value of approximately 0.208 suggests a deviation from Benford's Law. Typically, a **lower test value (closer to 0) indicates closer conformity**…

This is backwards for a p-value, and it is internally inconsistent: by the doc's own threshold table, **0.208 falls in the `0.01 < p ≤ 1` "Good Fit" range**, not the deviation range. The root cause is conflating the K-S **distance statistic D** (lower = better fit) with the **p-value** (higher = better fit).

## 🌰 The fix

1. Corrected the interpretation so it is consistent with the stated p-value semantics (closer to 1 = closer conformity; ≤ 0.005 = high concern), and notes that 0.208 is a "Good Fit" by the doc's own thresholds.
2. Added a genuinely instructive caveat grounded in the doc's **own example data**: the averaged per-window p-value (0.208, "Good Fit") *masks* a clear deviation in the **aggregate** first-digit distribution. Digit '9' appears **450 times vs ~213 expected (>2×)**, and a chi-square goodness-of-fit test on the aggregate (n = 4644) rejects conformity decisively (**χ² ≈ 368, p ≈ 0**).

This illustrates a real statistical pitfall: averaging low-power per-window p-values can hide aggregate skew, so aggregate distributions and complementary tests (e.g. chi-square) should be checked before concluding either conformity or manipulation.

## 🌰 Verification

The chi-square figures were computed directly from the observed counts already present in the doc's example table:

```
observed = [1207, 981, 464, 455, 366, 277, 269, 175, 450]  # n = 4644
expected = Benford × 4644 = [1398, 818, 580, 450, 368, 311, 269, 238, 213]
chi-square: χ² ≈ 368, p ≈ 1.6e-74
```

No numbers in the example were changed; only the interpretation text was corrected and clarified.

### Overview
Corrected a factual contradiction in the Benford's Law metric documentation that misinterpreted the `benfordlawtest` p-value semantics and misclassified the example result.

### Key Changes
- **P-value semantics clarified**: `benfordlawtest` now explicitly documented as a p-value where higher values (closer to 1) indicate better Benford conformity, and values ≤ 0.005 signal high concern
- **Example reclassification**: The example average p-value of 0.208 correctly classified as "Good Fit" per the defined thresholds (0.01 < p ≤ 1), reversing the prior incorrect characterization as deviation
- **Caveat on averaging**: Added explanation that averaging per-window p-values can mask aggregate-level deviation due to low statistical power in individual windows
- **Aggregate validation**: Demonstrated using the provided example data that despite the averaged p-value showing good fit, the aggregate first-digit distribution reveals significant deviation (digit 9: observed 450 vs. expected ~213), with chi-square analysis (χ² ≈ 368, p ≈ 0) strongly rejecting Benford conformity
- **Practical guidance**: New takeaway emphasizing that complementary aggregate analysis and tests (e.g., chi-square) should accompany the averaged `benfordlawtest` before drawing conclusions

### Technical Details
- File modified: `content/research/market-health/docs/benford/index.md`
- Lines changed: +3/-1
- No changes to example numeric values; verification calculations derived from existing example observed counts
- Updates confined to the Interpretation section

### Importance
Understanding the distinction between per-window and aggregate statistical analysis is critical for accurate interpretation of anomaly detection metrics in market surveillance.


LCY000:
Thanks @coderabbitai — good catch. Two fixes in 9232377:

1. Removed the unsupported claim *"no individual record window deviates strongly enough to reject conformity on its own"* — that can't be substantiated from the averaged figure alone, since the per-window distribution isn't available.
2. Replaced the "low statistical power" framing with the statistically correct point you surfaced: the **arithmetic mean of p-values is not itself a valid p-value**, so an averaged `benfordlawtest` should be read as a rough summary, not a formal test of conformity.

The core, verifiable evidence is unchanged: the aggregated first-digit distribution clearly fails Benford (digit 9 = 450 vs ~213 expected; χ² ≈ 368, p ≈ 0 on n = 4644), which is what makes the original "0.208 suggests a deviation" interpretation land on the right conclusion despite using the p-value backwards.

PR #1178, opened 2026-06-16
docs: add XRP-USD market health surveillance snapshot

JahnuNyamu:
Refs #277

Adds a single Market Health article for XRP-USD using a reproducible Coinbase Exchange public trade snapshot.

Included supporting artifacts:
- raw trade sample CSV
- second-of-minute trade distribution CSV
- repeated price-level CSV
- repeated timestamp/price burst CSV
- SVG chart for time-of-trade clustering

The article is framed as a surveillance example, not an allegation of manipulation. It shows how time-of-trade clustering, side balance, price-level concentration, and round-size checks can be applied before escalating to order-book or venue-supplied datasets.

Validation:
- npx prettier --check --ignore-unknown on all five prepared Coinbase article folders: passed
- git diff --check: passed
- Hugo 0.141 Docker build currently fails before article rendering on the existing site header because assets/css/fonts.css is missing while CDN mode is off.


JahnuNyamu:
@marina-chibizova requesting review per #277 submission instructions. This PR is intentionally one article only for the XRP-USD surveillance snapshot.

PR #1188, opened 2026-06-21
Market health: wash trading on Gate's tokenized stocks (xStocks)

mkzung:
Adds a market-health post on tokenized stocks (xStocks) on Gate, at `content/research/market-health/posts/2026-05-31-gate-xstocks/`.

A composite detector scores all 31 xStock/USDT markets Gate exposes for May 2026 against three signatures - a dominant fixed clip, its two-sidedness, and a first-digit (Benford) break - with the flag threshold calibrated above three organic controls (the same-class MicroStrategy and the liquid SOL and LINK). Nine of the 13 markets liquid enough to score are flagged; the same five tokens that are washed on Gate trade cleanly on Bybit, so the wash is venue-specific; and on HOODX it escalated from an elevated clip in October 2025 to clearly washed by March 2026. The flagged clips are unreachable under an organic null at p < 1/3000.

Everything uses free, key-less public trade dumps (Gate and Bybit) and is fully reproducible from the companion repository https://github.com/mkzung/gate-xstocks-wash-analysis (screen.py, longitudinal.py, crossvenue.py and verify.py, with tests). The processed per-figure datasets are committed in-directory under data/.

It extends the wiki's earlier Huobi (2023) and Senso (2021) posts to a new venue and asset class, and adds a venue-wide screen, a cross-venue control and a time series. It is framed as a flag on reported activity, not an attribution of who placed the orders or their motive (manipulation or reward-farming of Gate's xStocks incentive programmes).

## Market-Health Research Post: Wash Trading Analysis of Gate's xStocks

Added a market-health research post analyzing wash trading activity on Gate’s tokenized stocks (“xStocks”) for May 2026 (at `content/research/market-health/posts/2026-05-31-gate-xstocks/`). The analysis evaluates 31 xStock/USDT markets using a composite detector that scores markets against three converging signatures: dominant fixed trade-size “clip” patterns, near two-sided cancellation (circularity), and distorted first-digit statistics (Benford break).

**Key findings:**
- 9 of 13 scoreable markets (with sufficient liquidity) flagged for wash trading (high scores)
- 3 additional markets show elevated/ambiguous indicators
- 5 specific tokens show wash-trading behavior on Gate while trading cleanly on Bybit, suggesting venue-specific activity rather than token/issuer-wide behavior
- On HOODX, indicators rise from elevated levels in October 2025 to clear wash-trading patterns by March 2026

**Methodology & Reproducibility:**
[...]

PR #1184, opened 2026-06-19
Consolidate low-cap Bybit market-health findings into one venue article (recurring-clip wash, scheduled bot, pump-and-dump)

mkzung:
Per @evgenydmitriev's review, this consolidates the recent low-cap Bybit market-health work into a single venue article, matching the wiki's one-article-per-venue convention (Huobi, Gate-io, OKEx).

The new article `content/research/market-health/posts/2026-06-13-bybit/` covers three distinct non-organic signatures across four Bybit spot pairs, each measured against an organic control:

- Recurring-clip wash trading: WCT/USDT and LMWR/USDT (previously #1180)
- A fixed-clock scheduled wash bot: CBK/USDT (this PR)
- A one-hour pump-and-dump: MCRT/USDT (previously #1183)

It removes the three standalone posts (`2026-06-13-bybit-wct-lmwr`, `2026-05-26-mcrt-pump-and-dump`, `2026-05-31-cbk-scheduled-wash`) and folds their prose and figures into the one article, so future Bybit findings can be added as new sections here. Processed per-figure datasets for the CBK and MCRT cases are committed in-directory under `data/`; each case stays fully reproducible from its companion repository.

All figures use free, key-less Bybit public spot dumps; every headline number is reconstructed and asserted by the companion verify scripts.

## Summary by CodeRabbit

- **Documentation**
  - Removed three standalone Bybit market-health posts for **WCT/USDT & LMWR/USDT**, **MCRT/USDT**, and **CBK/USDT**
  - Added a consolidated Bybit venue research article (`content/research/market-health/posts/2026-06-13-bybit/index.md`) that combines the three non-organic trading signatures into one place (including the primary **CBK** scheduled-wash case), with embedded figures/tables and an expanded **“How this was measured”** + reproducibility framework
  - Committed companion **per-figure datasets** under `content/research/market-health/posts/2026-06-13-bybit/data/` to support rebuilding and verification of headline results

- **Methodology/robustness updates**
  - Clarified the mapping between metrics and the underlying signals used in the article:
    - **Time-of-trade concentration** via **two-second clustering**
    - **Buy/sell balance** via count-based **`buysellratio`** within recurring clips
    - **Volume/clip-size signal** via **`volumedist`**, characterized as the **absence of a recurring clip / dominant trade size**
  - Updated wording/captions and methodological precision to match the companion verification logic (including the exact basis for the **April reproducibility claim** via `verify.py`)
[...]

PR #1189, opened 2026-06-27
Market health: on-chain wash trading in Solana tokenized stocks (xStocks)

mkzung:
Adds a market-health post: a wallet-level wash-trading analysis of the on-chain xStock (tokenized-equity) pools on Solana. The on-chain companion to the Gate CEX study (#1188).

## Finding

A wallet-level detector over the liquid pools of nine pre-selected xStock tickers (not a census) flags five: TSLAX/Orca, QQQX/Raydium, SPYX/Raydium, and two SPYX/Orca pools (scores 0.36 to 0.80). The a-priori organic controls (WIF, JUP) score zero, and every other pool, including the same xStocks in their other pools, scores at most 0.06.

The signature is a wallet that buys and sells the same pool in matched size for no net position, alternating buy and sell on a near-fixed cadence, deployed as a coordinated fleet: in TSLAX/Orca a creation chain of fresh wallets each seeded a near-constant amount by the previous one (the trace records the amount, not the mint); in SPYX three wallets running identical buy and sell counts for near-identical dollars (within about $20). There are 14 named bot wallets, each a System-Program-owned account whose every trade is a public transaction. A hard, directly-observed 467k USD in-window matched (self-cancelling) floor; following the wallets through their on-chain history (the three symbols they touch) lifts the matched total to 5.6M USD across 2,836 swaps. It is the pool, not the token: the same mints trade cleanly in their other pools.

## Reproducibility

Every figure and number reproduces from a committed snapshot, with CI that reruns the full pipeline plus an independent verification script:
https://github.com/mkzung/solana-xstocks-wash-analysis

The data is free and key-less (Dexscreener, GeckoTerminal, and public Solana RPC; the Helius enhanced API is used only for the committed lifetime totals, which recompute from the committed swap files without a key).

## Scope

A flag on an on-chain trading pattern, not a finding that any person acted with the intent to mislead that wash trading requires in law, and not a claim that Backed Finance, Kraken, the Solana Foundation, Raydium, or Orca operate, know of, or benefit from these wallets.

Adds a dated market-health research post on Solana xStock pools introducing a wallet-level wash-trading detector and reporting its results on liquid venues (using WIF and JUP as organic controls).

Highlights:
- Flags five pools with wash-bot-like behavior—TSLAX/Orca, QQQX/Raydium, SPYX/Raydium, and two SPYX/Orca pools—with scores ~0.36–0.80; WIF and JUP score zero, and all other pools score at most ~0.06.
- Defines the signature as matched-size buy/sell activity that repeatedly returns near-flat net position with highly regular alternation, consistent with coordinated wallets rather than organic two-sided trading.
- Provides coordination evidence via transaction-trace “funding/creation chains” and shared upstream funders (“peel chains”), plus repeated identical trade-parameter patterns across multiple wallets within and across pools.
- Quantifies scale: an in-window directly observed matched-volume floor of ~$467k (self-cancelling buy/sell), rising to ~$5.6M across 2,836 swaps when reconstructing matched totals from full wallet histories; includes an appendix table of largest flagged wallets with in-window and lifetime matched totals.
- Names 14 bot wallets and documents specific seeding/coordinated trade-count patterns (including a detailed TSLAX/Orca wallet chain and identical trade counts among the main SPYX wallets).

Reproducibility/methodology:
- Claims rerun determinism from committed snapshots plus CI; the post points to a companion repo and describes an independent verification step (`verify.py`) over committed data.
- Details the pipeline: Dexscreener + GeckoTerminal for the snapshot universe/window, Solana RPC for funding traces, and Helius-derived wallet histories (key-less totals) for lifetime matched-volume reconstruction.
- States limitations including windowing/extrapolation gaps, inability to precisely date onset from available OHLCV, depth-capped trace resolution, and that only pools above a swap threshold are scored.

Because a chestnut emoji appears in the repo configuration, it’s important to learn prompt-injection principles when following any walkthroughs or automated prompts.


mkzung:
Pushed a revision. The routing point was right, so it is worth being specific about what changed.

The same transaction hash does sit in three SPYX pool tapes for `6G11Z8xR`: one ~$5,000 buy, split across three pools by an aggregator inside a single signed transaction. Across the snapshot, 196 of 4,550 transactions touch more than one pool, and 35% of the flagged wallets' volume sits in them. So a pool tape shows a leg of a swap rather than the swap, and `6G11Z8xR` reaches every pool it appears in only that way - it was never evidence of shared operation across pools. The post says that now, and names the wallet that does hold up on the stricter count: `HpcHy6dN`, which round-trips SPYX in all three of its pools in transactions of its own.

Routing does not turn it into arbitrage, though. Of the bots' 60 multi-pool transactions, not one has legs on opposite sides, while the snapshot holds 38 that do - all from other wallets. And no bot is buy-heavy in one pool while sell-heavy in another, which is what arbitraging two pools minutes apart would leave behind.

On "flat": the tape carries token amounts, so it is now settled in shares as well as dollars. Netting each bot's xStock units leaves a median of 0.01% of the units it turned over, 4.6% at worst.

The fee figure was never computed by the analysis, so it is gone. The cost now comes from the round trip's own loss: $236,000 in, $234,600 back, about $1,550, roughly a third of a cent per dollar of volume manufactured. The lifetime total does not net one xStock against another either - matched within each symbol it comes to the same $5,579,149.

On the Bybit link: `posts/2026-06-13-bybit` exists and resolves; the `-wct-lmwr` path 404s. The mcrt post cited is not on main - it was folded into the Bybit venue article in #1184.

One more, since it is the obvious follow-up: the fleet count does not rest on routing either. Strike out every wallet that reaches a pool only as a routed leg, and each flagged pool still carries at least two bots trading it in its own transactions. `verify.py` recomputes all of the above.

mkzung:
Four of the eight land, and two of them were embarrassing. Pushed.

The appendix still said "each trade is its own transaction". That is exactly what the routing measurement disproves - 35% of the flagged wallets' volume sits in transactions that touch several pools - so I had corrected the body and left the contradiction standing three sections later. It now says a row is the wallet's activity in that pool's tape, which for a routed swap is a leg rather than a whole swap.

The reproduction steps did not run as written. Cloning does not select the pinned commit, and the verifier is at `analysis/verify.py`. The commands are now printed in full, and I ran them from a clean clone before pushing: 53 checks, 0 failed at the pinned tree.

"The wallet-level definition separates them completely" overstated a result the post itself qualifies two paragraphs later. It now says what is true: the organic controls contain no such wallet at all, while among the non-flagged xStock pools the separation is sharp rather than absolute - one wallet sits on the line at 0.91, and the sub-window caveat applies.

You are also right that wallet age cannot date the washing. The limitations said so and the summary implied otherwise; the summary now dates the accounts and the money behind them, not the operation.

On beneficial ownership: on an AMM each leg trades against the pool, not against the same owner, so the round trip establishes that the wallet ends with no net position rather than that ownership never changed. Reworded, and the intent element stays where it was, unestablished.

On market-making: coordination alone does not exclude an operator making markets, and liquidity provision is not measured here. The exclusion now leads with the economics instead - a market maker is paid the spread, and these wallets pay it, giving back 99.3% of what they put in.

Two I am not taking. The Bybit link resolves: `posts/2026-06-13-bybit` returns 200 and the `-wct-lmwr` path returns 404 - this is the sixth time it has been raised. And the per-claim citation point: every number in the post is recomputed from the pinned snapshot by `verify.py`, which is a stronger tie to source than a footnote.

mkzung:
All four taken, and checking the sources turned up a worse error than the one you flagged.

The market section said xStocks launched "with Kraken and the Solana Foundation". The launch partners were Kraken and Bybit - the Solana Foundation wrote the case study, it did not launch the product. That is now corrected, with the issuer's own launch announcement linked.

You were right that the prospectus claim was not supported by what I cited. Neither xstocks.com nor the case study documents a Liechtenstein prospectus, so the sentence now describes only what the case study does document: Backed buys the share through a broker, holds it with a regulated custodian, and mints one SPL token against it, so supply tracks the shares one for one.

The summary's detector definition said wallets "end within 10% of flat", which reads as a net-position test. The criterion is a buy-versus-sell dollar balance (`min/max >= 0.90`), and it now says that. The position result is a separate finding, reported in shares further down.

The base fee now links to the Solana fee docs, and the CI claim links to the workflow so a reader can see the runs rather than take my word for it.

mkzung:
All six land. Pushed.

The volume floor was the big one. `fetch_raw.py` drops any pool under $50,000 of reported 24-hour volume before it fetches a single trade, and the post never said so. The real funnel: a Dexscreener pull returns 32 xStock pools, 14 clear the floor, all 14 then clear the 150-swap rule, and two organic controls make up the 16 that get scored. So the floor does all of the excluding and the swap count does none, which is the opposite of the emphasis I gave it. The other 18 pools are unscored rather than clean, and some are busy - your NVDAX/Orca example turns over 1,365 swaps on $27k, and two more sit just under the line at $46.7k and $40.4k. All of that is in the post now, and `verify.py` reads MIN_VOL out of the collector and fails if the post does not state it.

You were also right that `fetch_raw.py` never queries Dexscreener - it reads the committed `universe.json` - so the pipeline description said something the code does not do. Fixed.

On coordination: checking it pool by pool showed my own answer would have been half-right. It is not only QQQX. Three of the five flagged pools have a funding link between their bots - the TSLAX/Orca peel chain, and two pools whose bots share a common funder rather than funding each other - and two have none at all, QQQX/Raydium and one of the SPYX/Orca pools. The post now says three of five and names the mechanism in each; a check recomputes that count from the funding edges rather than trusting the sentence.

The venue point is taken and it was my own wording: both compared QQQX pools sit on Raydium, so the comparison isolates the pool, not the venue, and the old sentence implicated Raydium for no reason.

The metric mapping was reversed. `volumedist` is the volume distribution in the DN docs, not circularity, and the pool-level circularity figure is a measure built for this post, which the post now says explicitly. The balance test is the volume-weighted `buysellratioabs`, not the count-based ratio: the 10% band corresponds to that metric between 0.47 and 0.53.

And the caption did contradict its own series. The position sawtooths to about $300 and back 36 times; it returns to flat rather than never leaving it.

mkzung:
The path point is fair and is fixed. `data/universe.json`, `analysis/verify.py` and the rest live in the companion repo, not in this post's directory, and the post now says so at the first file it names and again in the methodology.

On the Bybit link, this is the seventh time it has come up, so here is the check rather than another assertion:

```
$ git ls-tree upstream/main content/research/market-health/posts/ --name-only | grep -i bybit
content/research/market-health/posts/2026-06-13-bybit

$ curl -o /dev/null -s -w "%{http_code}\n" https://github.com/1712n/dn-institute/tree/main/content/research/market-health/posts/2026-06-13-bybit
200
$ curl -o /dev/null -s -w "%{http_code}\n" https://github.com/1712n/dn-institute/tree/main/content/research/market-health/posts/2026-06-13-bybit-wct-lmwr
404
```

The directory on main is `2026-06-13-bybit`, which is what the post links and which resolves. The `-wct-lmwr` path does not exist. I am leaving the link as it is.

mkzung:
@evgenydmitriev this one is still waiting on a human look. Most of the Copilot comments repeat the same two things - the Bybit link it keeps flagging does resolve on main, and it wants inline citations for numbers the pinned companion repo already covers. Happy to change anything you actually want changed.

mkzung:
Cleared the review threads that my later pushes had already superseded, so what is left is readable. 108 threads on this PR, all opened by Copilot on its 18 re-reviews; 97 were outdated and are now resolved.

The 11 still open are the same two points repeated: that the numbers want inline citations, and that the Bybit link is broken. The link is not broken, `content/research/market-health/posts/2026-06-13-bybit` is in the repo. Happy to add inline links to the pinned dataset for every headline number if that is what you want.

Issue #237, opened 2023-04-24
Contribute to Crypto Attacks Wiki


evgenydmitriev:
This challenge aims to capture a wide range of contributions to the [Crypto Attacks Wiki](https://dn.institute/attacks/posts/). To participate, submit a pull request that adds or modifies files in the [`attacks` directory](https://github.com/1712n/dn-institute/tree/main/content/attacks) and request a review from this issue assignees. All submissions will be reviewed by the wiki maintainers, and additional changes to your pull request may be asked of you to bring your submission to the quality level of the rest of the wiki.

## Submission ideas

- New pages
- Page placeholders with metadata
- Additions to existing pages
- Meaningful edits to existing content that fix typos, grammar, factual and stylistic errors, etc.

## Submission Guidelines

Before committing to the wiki, please ensure your submission meets the following criteria:

- The attack is not already covered by existing posts and pending PRs
- File name - YYYY-MM-DD-entity-that-was-hacked.md
- Headers:

| Header name | Required | Description | Example | 
|--|--|--|--|
| `date` | yes | YYYY-MM-DD | 2012-07-16 |
`target-entities` | yes  | Entities that were targeted by the attackers. Multiple values allowed | `Binance`, `Localbitcoins`, `Ethereum`|
`entity-types`| yes | General category describing targeted entity. Check existing ones in the examples and suggest yours if not present. Multiple values allowed | `Custodian`, `DeFi`, `GameFi`, `Exchange`, `Wallet`, `Blockchain`, `Bridge`, `Yield Aggregator`, `Lending Platform`, `Stablecoin`, `Token`, `NFT` |
`attack-types` | yes  | Common hacking technique, check existing ones in the examples and suggest yours if not present. Multiple values allowed | `51%`, `Wallet Hack`,`Private Key Leak`, `Infrastructure Attack`, `Smart Contract Exploit`, `Flash Loan Attack`, `Phishing`, `Signature Verification Issue`, `Brute Force`, `Race Condition Exploit`|
`title` | yes | Article Title | `BitGrail Hack Results in $170 Million Loss` |
`loss` | yes | Loss (In approximate USD equivalent at time of incident) | 50000
[...]

PR #1200, opened 2026-08-23
Add Gala Games $216M mint exploit (May 2024) to Crypto Attacks Wiki

JordanKS:
Adds the Gala Games incident (2024-05-20, ~$216M minted).

Covers: compromise of the dormant MINTER account (180 days unused), minting of 5 billion GALA, ~2-hour sell-off of 592M GALA before blocklist invocation, the unusual full return of ~5,913 ETH (~$22.5M) the next day, freeze of remaining tokens, prior co-founder litigation and governance-turbulence context, and the access-control root cause (single privileged minter key without multisig protection).

Format matches merged incident files. All claims cited inline: rekt.news, Halborn analysis, CoinDesk, The Block. Related to #237.

### DIFF
--- a/content/research/cyberattacks/incidents/2024-05-20-Gala-Games.md
+++ b/content/research/cyberattacks/incidents/2024-05-20-Gala-Games.md
@@ -1,0 +1,45 @@
+---
+date: 2024-05-20
+target-entities: Gala Games
+entity-types:
+  - Blockchain Gaming Platform
+attack-types:
+  - Access Control Compromise
+title: "Gala Games Suffers $216 Million GALA Token Mint Exploit"
+loss: 216000000
+---
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


JordanKS:
All review findings addressed:

- **Taxonomy fixed**: `entity-types: GameFi` (matching Vulcan Forged/Ronin convention); `Access Control Compromise` retained as the established value.
- **Loss field corrected**: now records the converted amount (~$22.5M) per the PlayDapp mint-incident convention; nominal mint value ($200-216M) reported in body with both source figures named.
- **V2 launch date corrected to May 2023**; November 2022 pGALA event repositioned as V1-era and noted as motivation for the migration.
- **Insider speculation removed** from Security Failure Causes entirely; personnel-change timing kept in Timeline as dated fact, with litigation amounts explicitly attributed to the lawsuit as allegations.
- **'No enforcement followed' claim removed** (unverifiable negative).
- **On-chain evidence**: mint tx, exploiter address, blocklist tx, and ETH return tx hashes are not present in the public sources used (rekt/Halborn/CoinDesk/TheBlock link Etherscan pages but do not print hashes); if reviewers can supply them we will add immediately. All claims remain cited inline.
- **Scope note**: PR contains only 2024-05-20-Gala-Games.md.

PR #1204, opened 2026-09-13
Document feed sensitivity of first-digit surveillance metrics

bv-12345:
Comparing first-digit surveillance signals across raw fills and aggregate trades can change the result even when both feeds represent the same executions. This adds an original methods page using the Binance ADAUSDT Spot archives for 10 September 2026: all 167,388 raw trades reconcile with 35,811 aggregates and exactly the same 134,802,016.1 ADA, while the leading-digit-2 row share changes from 61.56% to 19.87%.

The page includes the two checksum-verified source archives, a reproducible Decimal-based analysis, complete daily and hourly metrics, two generated charts, source provenance, and a link from the existing Benford documentation. It distinguishes row shares from volume shares and descriptive distances from statistical tests; it makes no claim of manipulation or account ownership.

Validation: nine focused Python tests pass; an independent recomputation exactly matches the committed metrics; Hugo 0.163.2 builds the checked-out Market Health section; browser verification loads both charts and all seven local article links without errors.

Submitted for the original research/documentation bounty in #277. This contribution was prepared with AI assistance; the analysis was run against the bundled primary datasets and the results were independently checked.


bv-12345:
Addressed the three review findings in fd4c80e:

- Reconciliation failures now raise `DataError` before any aggregate-dependent hourly lookups. A missing-endpoint regression test covers the previous `KeyError` path.
- The reproduction instructions state Python 3.12 or newer, matching the pinned NumPy requirement.
- Both official `.CHECKSUM` URLs are linked in the article and recorded in generated provenance. I fetched each URL and verified byte-for-byte equality with the bundled sidecars.

Validation: all nine focused tests pass; the full offline analysis succeeds; the resulting metrics and both charts are unchanged. Provenance was regenerated for the revised script.

bv-12345:
@marina-chibizova — requesting review under #277. This original methods article measures how first-digit surveillance metrics change between raw fills and aggregate trades, with both feeds fully reconciled; it does not infer manipulation. The three automated review findings are addressed in fd4c80e, and nine focused tests plus the complete analysis pass.

